PR #16, opened 2016-04-08
Add a QR code

dag-io:
This PR export a new environment for a QR Code image url to the public install page.
This is particularly useful if in a later step you create a message with the download link + the QR code. Then if the tester view the message on his computer, he can grab his mobile and scan the QR code and voila.

Please review carefully as it's my 1st real PR for Bitrise. It might not be perfected. It has been tested via "bitrise run test" and with a sandbox project


viktorbenei:
This is a really good idea, definitely worth to implement!

Just one thing, could you please create a feature request on [UserVoice](https://bitrise.uservoice.com) instead? There others can vote on the request too and this feature will be implemented in the website, not in this Step.

dag-io:
Thanks @viktorbenei. There is actually already one here : https://bitrise.uservoice.com/forums/235233-general/suggestions/11024070-show-qr-code-in-public-install-page ;-)

viktorbenei:
You're right, thank you!
I'll close this issue then, and we'll keep track of the request on UserVoice :)

PR #18, opened 2016-06-20
filter package infos & tests

godrei:


spec/uploaders/apk_uploader_spec.rb
```diff
@@ -0,0 +1,31 @@
+require_relative './../../uploaders/apk_uploader'
+
+describe 'apk_uploader' do
+  describe 'filters test' do
+    it 'it filters package infos' do
```
viktorbenei: no need for second `it`

PR #27, opened 2017-02-17
Check emails format

trapacska:


step.rb
```diff
@@ -14,6 +14,17 @@ def fail_with_message(message)
   exit(1)
 end
 
+def validate_emails(emails)
+  if !emails.nil?
```
godrei: if emails.to_s == '' 
  return
end

Issue #22, opened 2016-09-05
Public install page for Xamarin builds


bykovme:
Xamarin build produces both APK & IPA file and there are 2 separate install pages for the files generated. Nevertheless, there is only one variable for public installation page ($BITRISE_PUBLIC_INSTALL_PAGE_URL ) in the output and only one link to installation page is assigned there  (points to APK in my case), while I need on the next steps to know the links to both installation pages to inform the testers in slack.
Could you produce additional separate variables for APK & IPA on this step?

viktorbenei:
Thanks for the feature request!

Right now you can do this by using multiple Deploy to Bitrise.io steps, pointing it to the IPA / APK directly, instead of to the directory where the files are stored.

In the future we plan to add a new environment variable which will include a list of the public install pages, instead of just one item.

bykovme:
Thanks for the tip with multiple deploy steps and looking forward for the new version of the step :)

freever:
Yeah the multiple deploy steps option is a good one!

viktorbenei:
A new option will be available, introduced in https://github.com/bitrise-io/steps-deploy-to-bitrise-io/pull/54 ; the `BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP` output will include a map of all the generated public install pages.

Not released yet, but should be soon (after merging https://github.com/bitrise-io/steps-deploy-to-bitrise-io/pull/55 & publishing a new version) :)

PR #54, opened 2017-11-29
Export all public URLs to the environment

### LINKED ISSUE #43: Public Install Page for multiple artifacts
When building for multiple platforms at the same time, this step will copy all artifacts available and create individual pages for each one, but it will set BITRISE_PUBLIC_INSTALL_PAGE_URL for just one of the generated artifacts.

I believe this should be changed to set a BITRISE_PUBLIC_INSTALL_PAGE_URLS (plural) with a comma separated list of the generated pages or make the install page support multiple platforms at once (which is the best option, but it implies changes in the bitrise platform...).

If you agree with the first option, I'd volunteer to try implement it.

philippelatulippe:
To support workflows where multiple artifacts have an install page, I've had this step export two new environment variables that contain all public URLs, instead of just the last one. The URLs are accessible through:
* a pipe-separated array of URLs, `BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST`
* a map of files and URLS, `BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP.` Entries are separated by pipes (|), keys and values a colon (:)

This fixes #43 and #22

### DIFF
--- a/main.go
+++ b/main.go
@@ -229,21 +229,14 @@ func main() {
 			fail("Failed to export BITRISE_PUBLIC_INSTALL_PAGE_URL, error: %s", err)
 		}
 		log.Printf("The public install page url is now available in the Environment Variable: BITRISE_PUBLIC_INSTALL_PAGE_URL (value: %s)", publicInstallPage)
-	}
-
-	if len(publicInstallPages) > 0 {
-		value := strings.Join(publicInstallPages, "|")
-		if err := tools.ExportEnvironmentWithEnvman("BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST", value); err != nil {
-			fail("Failed to export BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST, error: %s", err)
-		}
-		log.Printf("An array of public install page urls is now available in the Environment Variable: BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST (value: %s)", value)
+		log.Printf("")
 	}
 
 	if len(publicInstallPageMap) > 0 {
 		tuples := make([]string, 0)
 
 		for file, url := range publicInstallPageMap {
-			tuples = append(tuples, file + ":" + url)
+			tuples = append(tuples, file + "," + url)
 		}
 
 		value := strings.Join(tuples, "|")
@@ -251,6 +244,7 @@ func main() {
 			fail("Failed to export BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP, error: %s", err)
 		}
 		log.Printf("A map of deployed files and their public install page urls is now available in the Environment Variable: BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP (value: %s)", value)
+		log.Printf("")
 	}
 
 	// --
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/bitrise-io/go-utils/log"
 	"github.com/bitrise-io/go-utils/pathutil"
@@ -166,6 +167,8 @@ func main() {
 	log.Infof("Deploying files")
 
 	publicInstallPage := ""
+	publicInstallPages := make([]string,0)
+	publicInstallPageMap := make(map[string]string)
 
 	for _, pth := range clearedFilesToDeploy {
 		ext := filepath.Ext(pth)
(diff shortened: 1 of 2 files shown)


philippelatulippe:
Using a colon (:) as the separator between key and value was a bad idea considering the value will always contain a colon. I'm not sure what would be best, an asterisk, an exclamation point, the greater than sign.

kdobmayer:
Hi @philippelatulippe !

Thanks for the PR. Do we really need the `BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST` variable? The map seems to be enough and we can parse the url list and files from it.

The separator between keys and values could be a comma `,` or a colon followed by a space. But a simple colon is not bad, just split the value at the first colon to get the file and the url.

main.go
```diff
@@ -221,6 +230,29 @@ func main() {
 		}
 		log.Printf("The public install page url is now available in the Environment Variable: BITRISE_PUBLIC_INSTALL_PAGE_URL (value: %s)", publicInstallPage)
 	}
+
+	if len(publicInstallPages) > 0 {
+		value := strings.Join(publicInstallPages, "|")
+		if err := tools.ExportEnvironmentWithEnvman("BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST", value); err != nil {
+			fail("Failed to export BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST, error: %s", err)
+		}
+		log.Printf("An array of public install page urls is now available in the Environment Variable: BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST (value: %s)", value)
```
kdobmayer: Put a newline after logging the env vars to be more readable.

philippelatulippe: Do you mean to print a newline to the log or to put a newline in the code?

philippelatulippe:
Thanks for reviewing my merge request. :) I kept `BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST` to make things simple for people who just want an unordered list of URLs and don't care which is which. It's particularly nice for people using shell scripts, which has poor string processing facilities. I wouldn't mind removing it though, it would simplify the code.

That's also why I agonized over the key/value separator. I'll switch it to a comma, it seems ideal.

viktorbenei:
I agree @philippelatulippe re the `_LIST`, that it would be great for Bash scripts where string manipulation might be a problem, but even then it shouldn't be just a list of the URLs IMO as in that case there's absolutely no indication which URL is for which artifact. E.g. sending the list to Slack, no one would know which URL is what.

The map solution seems to fit this better, at least for now. Won't be "that nice" on slack but it's manageable I think.

Or, alternatively the `_LIST` one could also be something like: `filename1:URL1|filename2:URL2`, so basically a simplified version of the map.

viktorbenei:
My bad, I thought the `_MAP` is in JSON, but checked the code again and it's exactly what I described :D

So in that case I'd +1 to only have the `_MAP`, as the `_LIST` with only the URLs seem to have limited utility.

philippelatulippe:
I've fixed the issues from the review:
* Don't export an array of files, only the map
* print a newline after logging the value

I've also used a comma (,) as the key/value separator, since it makes it easier to use a split function to get the key and value.

philippelatulippe:
I accidentally pushed an old version of my code, I've now amended the commit.

kdobmayer:
You can iterate over the public page urls with the following code snippet:

```shell
# example apk paths and urls
BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP="/path/file.apk,https://url.com|/path/file2.apk,https://url2.com"

while read -r line; do
  echo "public page url: ${line#*,}"
done <<< "$(echo "${BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP}" | tr '|' '\n')"
```

You can use any command instead of `echo`.
Or put all of them into an array and use it later:

```shell
BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST=()
while read -r line; do
  BITRISE_PUBLIC_INSTALL_PAGE_URL_LIST+=("${line#*,}")
done <<< "$(echo "${BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP}" | tr '|' '\n')"
```

It's also working with colon (`:`) as a separator between paths and urls.

kdobmayer:
Hi @philippelatulippe ,

We discussed it and decided to use only the map as output. I will merge your PR then remove the list output.

philippelatulippe:
I thought I had already removed the map output. Did I forget to push? Or
perhaps I forgot to update the yaml. I'm not currently at my laptop so I
can't fix it right now, I'd be glad if you merged and fixed it, but
otherwise I can fix it later this afternoon.

viktorbenei:
Merged & released, thanks for the PR @philippelatulippe ! ;)

PR #69, opened 2018-09-10
Fix platform build version name

BirmacherAkos:
# Description

- fix(regex): filterPackageInfos() regex fix: it will igonre the apk's "platformBuildVersionName" versionName after the versionName
- test(android): android test workflows added

Click Up: https://app.clickup.com/757956/762238/t/76qc3

## Type of change
Please select options that are relevant.
- [X] Issue
- [ ] Enhancement
- [ ] Maintenance
- [ ] Feture request


uploaders/apkuploader_test.go
```diff
@@ -0,0 +1,371 @@
...
+		name    string
+		aaptOut string
+		want    string
+		want1   string
+		want2   string
+	}{
+		{
+			name: "test_with_empty_platformBuildVersionName",
+			aaptOut: `package: name='com.example.birmachera.myapplication' versionCode='1' versionName='1.0' platformBuildVersionName=''
+		sdkVersion:'17'
```
godrei: please do not use the whole output for every test.
It is fine to test the whole output once, but other than this, just test the relevant line of the log (or some log from its environment).
Also in case of the whole output test, please define the whole output using a var, to make the tests more readable.

PR #74, opened 2018-09-20
Fix isEnablePublicPage argument in DeployFile

### LINKED ISSUE #73: Can't get install pages except for .apk and .ipa files
## Summery
- I would deploy mapping.txt file to bitrise with using this step.
- I could get public install page for my .apk files, however I couldn't get it for my mapping.txt
- I attach log below.
- I read source code and found the reason why.
log

## Reason
- When this step upload files except for .apk and .ipa files, it call [`DeployFile`](https://github.com/bitrise-io/steps-deploy-to-bitrise-io/blob/4699a3e2fec1cf72b8ac229b779c59b3220cac34/main.go#L218).
- In `DeployFile`, [`finishArtifact` will call with `"no"` argument.](https://github.com/bitrise-io/steps-deploy-to-bitrise-io/blob/4699a3e2fec1cf72b8ac229b779c59b3220cac34/uploaders/fileuploader.go#L16)
- [`finishArtifact` function always checks if `isEnablePublicPage` is `"true"`](https://github.com/bitrise-io/steps-deploy-to-bitrise-io/blob/4699a3e2fec1cf72b8ac229b779c59b3220cac34/uploaders/common.go#L211), but now `isEnablePublicPage` parameter is `"no"`. So, `finishArtifact` returns `"", nil` even if `artifactResponse.PublicInstallPageURL` is not empty.

## Question
- Why `finishArtifact`'s 7th argument in `DeployFile` is always `"no"` ?
- If it is mistake, I want to fix it.

ronnnnn:
## Summary
- fix #73 
- We can get public install page for all type files.

### DIFF
--- a/uploaders/fileuploader.go
+++ b/uploaders/fileuploader.go
@@ -13,7 +13,7 @@ func DeployFile(pth, buildURL, token, notifyUserGroups, notifyEmails, isEnablePu
 		return "", fmt.Errorf("failed to upload file artifact, error: %s", err)
 	}
 
-	publicInstallPage, err := finishArtifact(buildURL, token, artifactID, "", "", "", "no")
+	publicInstallPage, err := finishArtifact(buildURL, token, artifactID, "", "", "", isEnablePublicPage)
 	if err != nil {
 		return "", fmt.Errorf("failed to finish file artifact, error: %s", err)
 	}


ronnnnn:
close this #73

PR #79, opened 2019-02-05
initial test result converter interface

trapacska:
Waiting on the website team:
- no asset upload yet
- authentication will be changed: will be removed from the URL and will be adde din the req header. (https://github.com/bitrise-io/steps-deploy-to-bitrise-io/pull/79/files#diff-bb46479a150e4b41cce2d9d95828929bR56)
- now the step requires the addon to be turned on on the app's settings tab


test/converters/converters.go
```diff
@@ -0,0 +1,21 @@
+package converters
```
trapacska: This file contains the interface that is required to be a package a test result converter. It must be possible to set files from outside(for example if someone wants to use a pre-filtered files list), need to return Junit4 xml test result, and needs to have a Detect method to see if the converter can run with the files included in the test result dictionary. (So a converter can run only if the dir has a TestSummaries.plist file for example)

test/converters/junitxml/junitxml.go
```diff
@@ -0,0 +1,31 @@
+package junitxml
```
trapacska: Basic converter for android junit4 xmls

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
+package xcresult
```
trapacska: This converter is responsible for detecting *.xcresult files and parse it's TestSummaries.plist file, then convert it into a Junit4 xml format.

trapacska: Here you can see what structure needs to be parsed: https://github.com/bitrise-io/steps-deploy-to-bitrise-io/pull/79/files#diff-f1794a0dd4ef21c422ef14bd6f251acfR253

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/bitrise-io/bitrise/models"
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/steps-deploy-to-bitrise-io/test/converters"
+	"github.com/google/uuid"
+)
+
+func findImages(testDir string) (imageFilePaths []string) {
```
trapacska: So the supported image asset files will be:
*.png
*.PNG
*.jpg
*.JPG
*.jpeg
*.JPEG

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+type ResultAssets struct {
+	XMLContent []byte                    `json:"xml"`
+	ImagePaths []string                  `json:"-"`
+	StepInfo   models.TestResultStepInfo `json:"step_info"`
+}
+
+// Bundle ...
+type Bundle []ResultAssets
+
+// ParseTestResults ...
```
trapacska: Lists dirs from the rootTestsDir, finds all files from the next level then iterates through all the converters and to this dirs and appends the xml's and assets into a bundle if a converter is detected for a directory

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+					ImagePaths: findImages(filepath.Join(testsRootDir, testDir.Name())),
+					StepInfo:   *stepInfo,
+				})
+			}
+		}
+	}
+	return bundle, nil
+}
+
+func checkResponse(resp *http.Response) error {
```
trapacska: checks response code to be successful, if not it returns an error and if possible it returns the body content as well in the error

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+		bodyData, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("unsuccessful response code: %d and failed to read body, error: %s", resp.StatusCode, err)
+		}
+		return fmt.Errorf("unsuccessful response code: %d\nbody:\n%s", resp.StatusCode, bodyData)
+	}
+	return nil
+}
+
+// Upload ...
```
trapacska: Generates a map of custom IDs and file names. This is required to gather the same map from our backend but with upload urls.

So generates:
```
UUID1: filename1.png
UUID2: filename2.png
```
This being sent to the server and it responds an upload table:
```
UUID1: http://here-to-upload.com/file.php?file=filename1.png
UUID2: http://here-to-upload.com/file.php?file=filename2.png
```
Then the code uploads the files to the matching URLs.
If all the files are uploaded then we post a final request with the test result xml and the step info.

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+	Duration         float64
+	TestStatus       string
+	TestIdentifier   string
+	TestName         string
+	TestObjectClass  string
+	Subtests         Subtests
+	FailureSummaries []FailureSummary
+}
+
+func (st Subtest) Failure() (message string) {
```
BirmacherAkos: Note: Personally I would not use named return value in this case.

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+					StepInfo:   *stepInfo,
+				})
+			}
+		}
+	}
+	return bundle, nil
+}
+
+func checkResponse(resp *http.Response) error {
+	if resp.StatusCode < 200 || 299 < resp.StatusCode {
```
BirmacherAkos: What about:
`if resp.StatusCode < http.StatusOK || http. StatusIMUsed < resp.StatusCode`

More about it:
https://golang.org/pkg/net/http/

---

BTW: I think it will be enough to check for 
- StatusOK                   = 200
- StatusCreated           = 201

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
```
BirmacherAkos: I think this will shadow the other returned errors.

You should use a different error variable `cerr` like this:
```
defer func() {
		if cerr := response.Body.Close(); cerr != nil {
			log.Warnf("Failed to close response body, error: %s", cerr)
		}
	}()
```

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+		if err := resp.Body.Close(); err != nil {
+			fmt.Printf("Failed to close body, error: %s\n", err)
+		}
+	}()
+
+	if err := checkResponse(resp); err != nil {
+		return err
+	}
+
+	var assetUploadURLMap map[string]string
```
BirmacherAkos: I think you should use `make(map[string]string)` instead.

"Map types are reference types, like pointers or slices, and so the value of m above is nil; it doesn't point to an initialized map. A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic; don't do that."
https://blog.golang.org/go-maps-in-action

BirmacherAkos:
Appreciate the comments 👍 
Could you include them (most of them) in the go files too?

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+func (summaryPlist TestSummaryPlist) Tests() map[string]Subtests {
+	tests := map[string]Subtests{}
+	var subTests Subtests
+	for _, testableSummary := range summaryPlist.TestableSummaries {
+		for _, test := range testableSummary.Tests {
+			subTests = append(subTests, collapseSubtestTree(test.Subtests)...)
+		}
+	}
+	for _, test := range subTests {
+		testID := strings.Split(test.TestIdentifier, "/")[0]
```
lpusok: Why the first part? I think that is the project name, if that is/is'nt intended, it is not clear why.

test/converters/converters.go
lpusok: Package name is a bit too generic IMO.

test/test.go
trapacska: This is how you need to use for JSON unmarshal :) It is in the next line

test/converters/converters.go
godrei: it would be a great package level comment

test/converters/converters.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"github.com/bitrise-io/steps-deploy-to-bitrise-io/test/converters/junitxml"
+	"github.com/bitrise-io/steps-deploy-to-bitrise-io/test/converters/xcresult"
+)
+
+type Intf interface {
```
godrei: what about `Converter`

test/converters/converters.go
```diff
@@ -0,0 +1,21 @@
...
+	Detect() bool
+}
+
+var handlers = []Intf{
+	&junitxml.Handler{},
+	&xcresult.Handler{},
+}
+
+func List() []Intf {
+	return handlers
```
godrei: why do not you make the `handlers` an exported var?

test/converters/converters.go
lpusok: Supposedly this step is the main user of this interface. As long as there are 2 types (android, ios) this interface feels a bit complex for the intended use. It all depends how it is used in the client code, but it could be a function call instead of a for with 3 function calls. Let's discuss.

trapacska: 4+ converters are about to be implemented

test/converters/converters.go
trapacska: I want to keep this var unsettable from outside

test/converters/converters.go
trapacska: converters.Converter?

trapacska: also, the converter is that matching to the interface

test/converters/converters.go
trapacska: yeah, tried to be as generic as possible

test/converters/xcresult/xcresult.go
trapacska: This is how it is in the plist

trapacska: So this is why it is the "testsuite" name in our case

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+		prefix = "\n"
+	}
+	return
+}
+
+type Subtests []Subtest
+
+func (sts Subtests) FailuresCount() (count int) {
+	for _, test := range sts {
+		if len(test.FailureSummaries) > 0 {
```
lpusok: Is the goal to now to count failed tests or the number of FailureSummaries (not sure if 1 failing test has 1 FailureSummary)

trapacska: we print out all the failure summaries in one failure section per test case

trapacska: please see the sample files in the test file consts

test/converters/xcresult/xcresult.go
trapacska: message used only once, and only for appending stuff in it, this way there is a -1 line of declaration

test/test.go
trapacska: You are right, however not our team that writes the server to be able to use it. I was ware of the http constants however this way it is more easily visible that we are capturing status codes between an interval.

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+	testSummariesPlistPath string
+}
+
+func (h *Handler) SetFiles(files []string) {
+	h.files = files
+}
+func (h *Handler) Detect() bool {
+	for _, file := range h.files {
+		if strings.HasSuffix(file, ".xcresult") {
+			h.testSummariesPlistPath = filepath.Join(file, "TestSummaries.plist")
```
lpusok: Should be fine, good to know however that subfolders containing different build phases contain TestSummaries.plist version 1.2 with additional info about the running environment (os version, device etc.), instead of 1.1 in the root.

trapacska: Could you please explain this a bit more?

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	return tests
+}
+
+type TestableSummary struct {
+	TargetName      string
+	TestKind        string
+	TestName        string
+	TestObjectClass string
+	Tests           []Test
```
lpusok: can you move the Test/Tests definition closer?

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+			Name:     testID,
+			Tests:    len(tests),
+			Failures: tests.FailuresCount(),
+			Time:     tests.TotalTime(),
+		}
+
+		for _, test := range tests {
+			testSuite.TestCases = append(testSuite.TestCases, TestCase{
+				Name:      test.TestName,
+				ClassName: testID,
```
lpusok: Why is the ClassName and the TestSuite name testID too?

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (h *Handler) XML() ([]byte, error) {
+	data, err := fileutil.ReadBytesFromFile(h.testSummariesPlistPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var plistData TestSummaryPlist
+	if _, err := plist.Unmarshal(data, &plistData); err != nil {
```
lpusok: Nice

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+func (st Subtest) Failure() (message string) {
+	prefix := ""
+	for _, failure := range st.FailureSummaries {
+		message += fmt.Sprintf("%s%s:%d - %s", prefix, failure.FileName, failure.LineNumber, failure.Message)
+		prefix = "\n"
+	}
+	return
+}
+
+type Subtests []Subtest
```
lpusok: Is it needed?

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (sts Subtests) TotalTime() (time float64) {
+	for _, test := range sts {
+		time += test.Duration
+	}
+	return time
+}
+
+type Test struct {
+	Subtests Subtests
```
lpusok: Not sure about this one

gaborszakacs:
I would prefer this workflow:
1. Make a POST request with all the related info of the test report and get the upload URLs as a response.
2. Upload all the things.
3. Make a PATCH request indicating that all the things were uploaded.

Some additional notes:
- I would prefer having the XML as an uploaded file too.
- I would prefer having the filename as unique (you can include the path in the filename if needed.)

```
(You have to include the auth token in the header, the same one the FTL step uses.)

POST http://localhost:5000/test/apps/<app_slug>/builds/<build_slug>/test_reports
{
  "filename": "junit.xml",
  "filesize":3423423,
  "step": {
    // step info (I'm not sure what goes here.)
  },
  "assets": [
    {
      "filename":"img1.jpg", // this should be unique across assets in this test_report
      "filesize":3423
    },
    {
      "filename":"img2.jpg",
      "filesize":3423
    }
  ] 
}

response 201 OK
{
  "id": "98b44204-e337-4d5e-8848-e8a44015f131",
  "filename": "junit.xml",
  "upload_url": "https://storage.googleapis.com/..."
  "assets": [
    {
      "filename":"img1.jpg",
      "upload_url":"https://storage.googleapis.com/..."
    },
    {
      "filename":"img2.jpg",
      "upload_url":"https://storage.googleapis.com/..."
    }
  ] 
}

// Upload all the things to Google bucket


PATCH http://localhost:5000/test/apps/<app_slug>/builds/<build_slug>/test_reports/98b44204-e337-4d5e-8848-e8a44015f131
{"uploaded":true}

response 200 OK
```

test/converters/xcresult/xcresult.go
trapacska: It is because the xml unmarshal, this is the only way I could group xml elements as the standard junit4 xml.

test/converters/converters.go
```diff
@@ -0,0 +1,21 @@
...
+	XML() ([]byte, error)
+	Detect() bool
+}
+
+var handlers = []Intf{
+	&junitxml.Handler{},
+	&xcresult.Handler{},
+}
+
+func List() []Intf {
```
godrei: comments are missing go lint will fail

test/converters/converters.go
lpusok: Ok, what about 1 function then, XML([]string) ([]byte, error)

test/converters/xcresult/xcresult.go
lpusok: It's worth to add a comment. Consider testSuite instead of testID.

test/converters/xcresult/xcresult.go
lpusok: See [](https://michele.io/test-logs-in-xcode/). There are more than 1 plists, mostly similar content.

test/converters/converters.go
```diff
@@ -0,0 +1,21 @@
+package converters
+
+import (
+	"github.com/bitrise-io/steps-deploy-to-bitrise-io/test/converters/junitxml"
+	"github.com/bitrise-io/steps-deploy-to-bitrise-io/test/converters/xcresult"
+)
+
+type Intf interface {
+	SetFiles([]string)
```
godrei: i would remove this func, why does it worth to hold the files if detect may filters out a couple of them?
pass the files to detect

test/converters/converters.go
```diff
@@ -0,0 +1,21 @@
...
+	XML() ([]byte, error)
```
godrei: i would just convert to a shared test results model and return the model, that model could be parsed on the client side (as in case of every converter we are looking for the same XML format, described by the same model).
this would also lead to a better name, for example: `Convert/Parse...`
this way also the interface could be: `TestResultConverter`

test/converters/junitxml/junitxml.go
```diff
@@ -0,0 +1,31 @@
...
+		if strings.HasSuffix(file, ".xml") {
+			h.xmlFilePath = file
+			return true
+		}
+	}
+	return false
+}
+
+func (h *Handler) XML() ([]byte, error) {
+	return fileutil.ReadBytesFromFile(h.xmlFilePath)
```
godrei: parse it to our shared model, to filter out unknown/unused fields of the resports

test/converters/xcresult/xcresult.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+//
+////
+
+// Junit4 XML
+
+type Junit4XML struct {
```
godrei: Let's convert this to the shared TestResult model, the target for every upcoming converter

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		return fmt.Errorf("unsuccessful response code: %d\nbody:\n%s", resp.StatusCode, bodyData)
+	}
+	return nil
+}
+
+// Upload ...
+func (bundle Bundle) Upload(client *http.Client, endpointURL string) error {
+	// generate UUID->filename and UUID->filepath maps and fill them with the assets need to be uploaded
+	assetUploadRequestMap, assetIDPathMap := map[string]string{}, map[string]string{}
```
lpusok: Map name would be more readable if containing both the key and value e.g. assetIDToFilename

test/converters/xcresult/xcresult.go
godrei: right below TestSummaryPlist, there is the only one reference for this model

test/converters/xcresult/xcresult.go
godrei: please create tests for this package i do not have a testSummary.plist, hard to understand and review the conversion without that

godrei:
move the test to the appropriate converter package and use golden files: https://medium.com/soon-london/testing-with-golden-files-in-go-7fccc71c43d3

test_test.go file has 14015 lines

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := checkResponse(resp); err != nil {
+		return err
+	}
+
+	var assetUploadURLMap map[string]string
+	if err := json.NewDecoder(resp.Body).Decode(&assetUploadURLMap); err != nil {
+		return err
+	}
+
+	// upload files matching to heir UUIDs
```
lpusok: typo in heir

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+		return fmt.Errorf("unsuccessful response code: %d\nbody:\n%s", resp.StatusCode, bodyData)
+	}
+	return nil
+}
+
+// Upload ...
+func (bundle Bundle) Upload(client *http.Client, endpointURL string) error {
+	// generate UUID->filename and UUID->filepath maps and fill them with the assets need to be uploaded
+	assetUploadRequestMap, assetIDPathMap := map[string]string{}, map[string]string{}
+	for _, testResultAsset := range bundle {
```
lszucs: `asset` instead of `testResultAsset`?

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+		for _, path := range testResultAsset.ImagePaths {
+			assetUUID := uuid.New().String()
```
lszucs: `uuid` instead?

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+			assetUploadRequestMap[assetUUID] = filepath.Base(path)
+			assetIDPathMap[assetUUID] = path
+		}
+	}
+
+	// Get an uploadURL map from the server UUID->URL to upload
+	var assetUploadRequestMapData bytes.Buffer
```
lszucs: `data` instead? It is used right after, nowhere else.

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+		return err
+	}
+
+	var assetUploadURLMap map[string]string
+	if err := json.NewDecoder(resp.Body).Decode(&assetUploadURLMap); err != nil {
+		return err
+	}
+
+	// upload files matching to heir UUIDs
+	for uuid, uploadURL := range assetUploadURLMap {
```
lszucs: `url` instead of `uploadURL`?

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+		if err != nil {
+			return fmt.Errorf("unsuccessful response code: %d and failed to read body, error: %s", resp.StatusCode, err)
+		}
+		return fmt.Errorf("unsuccessful response code: %d\nbody:\n%s", resp.StatusCode, bodyData)
+	}
+	return nil
+}
+
+// Upload ...
+func (bundle Bundle) Upload(client *http.Client, endpointURL string) error {
```
lszucs: Function does lots of things. Consider splitting to smaller functions, for example along commented blocks and http requests. Many variable naming issues arise because of large common scope. Splitting would aid in testing too.

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+
+	}
+
+	// post a confirmation the assets are uploaded and send the xml and step info with it as well
+	var bundleData bytes.Buffer
+	if err := json.NewEncoder(&bundleData).Encode(bundle); err != nil {
+		return err
+	}
+
+	fmt.Println("ITT", string(bundleData.String()))
```
lszucs: Accidentally committed debugging message?

test/converters/converters.go
trapacska: AFAIK test folder is not locked at all by any Go convention.

test/test.go
trapacska: by this sample code it looks like it works a little bit differently than the expected:
```
package main

import "fmt"

func test() (err error) {
	err = fmt.Errorf("tomi")

	defer func() {
		if err := fmt.Errorf("nemtomi"); err != nil {
			fmt.Println("from defer:", err)
		}
	}()

	fmt.Println("outer scope:", err)
	return
}

func main() {
	fmt.Println("main:", test())
}
```

test/converters/xcresult/xcresult.go
trapacska: comment added

test/converters/converters.go
trapacska: The interface methods reduced to XML() and Detect(). I think now both of these describe the converter properly. XML() is a Go convention for typed outputs like String() etc..

trapacska: we need to separate the logic for finding supported converter(s) and if they support the current test result type then we need to be able to run them

test/converters/converters.go
trapacska: This is the most common name I could find for the collection of converters.

test/converters/xcresult/xcresult.go
trapacska: I see, however I think it is the main goal to list all test cases nevertheless in whichbuild phases they are.

test/converters/converters.go
trapacska: The interface methods reduced to XML() and Detect(). I think now both of these describe the converter properly. XML() is a Go convention for typed outputs like String() etc..

trapacska: There is a converter interface to match for each converter. I will remove Handler completely as it is pointed out it sounds misleading.

test/converters/xcresult/xcresult.go
trapacska: 👍

test/converters/xcresult/xcresult.go
trapacska: In the junit xml it is grouped like that, I think testsuites are grouped in the named teststuites branch

test/converters/xcresult/xcresult.go
trapacska: thx

test/converters/xcresult/xcresult.go
trapacska: This is how I could follow the relevant plist structure.

test/converters/converters.go
trapacska: I would keep the XML output as each platform has different way to produce an xml. Each platform has a different logic to build up the Junit XML structure.

trapacska: So I think this is why it needs to be handled individually.

test/converters/xcresult/xcresult.go
trapacska: This model cannot be converted to our shared model as it is a strict model to have for marshaling the data structure into the standard Junit XML format. This is a standard.

test/converters/junitxml/junitxml.go
trapacska: same as https://github.com/bitrise-io/steps-deploy-to-bitrise-io/pull/79#discussion_r257692466

test/test.go
trapacska: this part is moved out completely

test/converters/xcresult/xcresult.go
trapacska: there was tests for that already

test/test.go
trapacska: removed

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+	"strings"
+
+	"github.com/bitrise-io/bitrise/models"
+	"github.com/bitrise-io/go-utils/fileutil"
+	"github.com/bitrise-io/steps-deploy-to-bitrise-io/test/converters"
+	"github.com/google/uuid"
+)
+
+func findImages(testDir string) (imageFilePaths []string) {
+	for _, ext := range []string{".jpg", ".jpeg", ".png"} {
```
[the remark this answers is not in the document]
trapacska: this instance used only once, there is no need to have it centralized

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+	StepInfo   models.TestResultStepInfo `json:"step_info"`
+}
+
+// Bundle ...
+type Bundle []ResultAssets
+
+// ParseTestResults ...
+func ParseTestResults(testsRootDir string) (bundle Bundle, err error) {
+	// read dirs in base tests dir
+	testDirs, err := ioutil.ReadDir(testsRootDir)
```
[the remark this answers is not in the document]
trapacska: Makes obvious here, but maybe not after couple of other lines/scopes

test/test.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// find test results in each dir, skip if invalid test dir
+	for _, testDir := range testDirs {
+		testDirPath := filepath.Join(testsRootDir, testDir.Name())
```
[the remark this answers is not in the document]
trapacska: same as: https://github.com/bitrise-io/steps-deploy-to-bitrise-io/pull/79#discussion_r257693869

test/test.go
trapacska: this got removed completely

test/test.go
trapacska: removed

test/test.go
trapacska: removed

test/test.go
trapacska: re-worked completely

test/test.go
trapacska: sure, removed 😄

trapacska:
This PR needs to be updated to not only upload images from the root but recursively look for images.

test/test.go
```diff
@@ -102,61 +117,88 @@ func findImages(testDir string) (imageFilePaths []string) {
...
-				if err != nil {
-					return nil, err
-				}
-				if err := json.Unmarshal(stepInfoFileContent, &stepInfo); err != nil {
-					return nil, err
-				}
-			}
+		stepInfoFileContent, err := fileutil.ReadBytesFromFile(filepath.Join(testDirPath, "step-info.json"))
+		if err != nil {
+			continue
```
godrei: i think we should log a warning for debugging reasons here, and maybe above `testDir.IsDir` as well

test/test.go
```diff
@@ -213,7 +256,7 @@ func (results Results) Upload(apiToken, endpointBaseURL, appSlug, buildSlug stri
 			}
 		}
 
-		if err := httpCall(apiToken, http.MethodPatch, fmt.Sprintf("%s/apps/%s/builds/%s/test_reports/%s", endpointBaseURL, appSlug, buildSlug, uploadResponse.ID), nil, nil); err != nil {
+		if err := httpCall(apiToken, http.MethodPatch, fmt.Sprintf("%s/apps/%s/builds/%s/test_reports/%s", endpointBaseURL, appSlug, buildSlug, uploadResponse.ID), strings.NewReader(`{"uploaded":true}`), nil); err != nil {
```
godrei: what about creating a variable for the upload URL, i would be more readbale:

```
url := fmt.Sprintf("%s/apps/%s/builds/%s/test_reports/%s", endpointBaseURL, appSlug, buildSlug, uploadResponse.ID)
httpCall(apiToken, http.MethodPatch, url, strings.NewReader(`{"uploaded":true}`), nil)
```

test/converters/xcresult/testsummariesplist.go
```diff
@@ -0,0 +1,105 @@
...
+	"strings"
+)
+
+// TestSummaryPlist ...
+type TestSummaryPlist struct {
+	FormatVersion     string
+	TestableSummaries []TestableSummary
+}
+
+func collapseSubtestTree(data Subtests) (tests Subtests) {
```
BirmacherAkos: Note:
I think it's better to avoid using named return values if possible.

test/test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	if output != nil {
+		return json.NewDecoder(resp.Body).Decode(&output)
+	}
+	return nil
+}
+
+func findImages(testDir string) (imageFilePaths []string) {
+	for _, ext := range []string{".jpg", ".jpeg", ".png"} {
+		if paths, err := filepath.Glob(filepath.Join(testDir, "*"+ext)); err == nil {
```
BirmacherAkos: Do not forget, that the glob uses a pattern:
https://golang.org/pkg/path/filepath/#Match

Fix:
https://github.com/bitrise-io/go-xcode/blob/b3f7172f910693d60973a07b61db757aca81f778/utility/glob.go#L4

test/test.go
```diff
@@ -126,12 +129,14 @@ func ParseTestResults(testsRootDir string) (results Results, err error) {
 
 	// find test results in each dir, skip if invalid test dir
 	for _, testDir := range testDirs {
+		// <root_tests_dir>/<test_dir>
+		testDirPath := filepath.Join(testsRootDir, testDir.Name())
+
 		if !testDir.IsDir() {
+			log.Debugf(colorstring.Yellowf("%s is not a directory", testDirPath))
```
godrei: the step has no input for turning on debug log level, so either:
- wire in input for it
- use another log severity
- remove the debug logs

main.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	AppSlug                    string `env:"BITRISE_APP_SLUG,required"`
 	AddonAPIBaseURL            string `env:"addon_api_base_url,required"`
 	AddonAPIToken              string `env:"addon_api_token"`
+	DebugMode                  bool   `env:"debug_mode,required"`
```
BirmacherAkos: Change the required flag to opts[yes,no]

main.go
```diff
@@ -6,29 +6,36 @@ import (
...
-	NotifyEmailList            string
-	IsPublicPageEnabled        string
-	PublicInstallPageMapFormat string
+// Config ...
+type Config struct {
+	BuildURL                   string `env:"build_url,required"`
+	APIToken                   string `env:"build_api_token,required"`
+	IsCompress                 string `env:"is_compress,opt[true,false]"`
+	ZipName                    string `env:"zip_name"`
+	DeployPath                 string `env:"deploy_path,dir"`
```
usmonster: Maybe there's a bug here...? See #83.

PR #81, opened 2019-04-18
Add support to clear the deploy directory when finished/failed

sepperousseau:
Add option that will clear the directory when the deploy is finished or failed.
This way when you have multiple deploys you can still use the same directory.

Because if you archive 3 times and deploy 3 times. The old items are still in that directory.
I could change the folder for every archive. But I like this handy flag.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -51,6 +51,7 @@ workflows:
         - notify_user_groups: $NOTIFY_USER_GROUPS
         - notify_email_list: $NOTIFY_EMAIL_LIST
         - is_enable_public_page: "true"
+        - clear_directory: "true"
     - path::./:
         title: IOS Test compressed with default name
         run_if: true
--- a/main.go
+++ b/main.go
@@ -29,6 +29,7 @@ type ConfigsModel struct {
 	NotifyEmailList            string
 	IsPublicPageEnabled        string
 	PublicInstallPageMapFormat string
+	ClearDirectoryWhenFinished string
 }
 
 // PublicInstallPage ...
@@ -48,6 +49,7 @@ func createConfigsModelFromEnvs() ConfigsModel {
 		NotifyEmailList:            os.Getenv("notify_email_list"),
 		IsPublicPageEnabled:        os.Getenv("is_enable_public_page"),
 		PublicInstallPageMapFormat: os.Getenv("public_install_page_url_map_format"),
+		ClearDirectoryWhenFinished: os.Getenv("clear_directory"),
 	}
 }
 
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -86,10 +91,12 @@ func (configs ConfigsModel) print() {
 	log.Printf("- NotifyEmailList: %s", configs.NotifyEmailList)
 	log.Printf("- IsPublicPageEnabled: %s", configs.IsPublicPageEnabled)
 	log.Printf("- PublicInstallPageMapFormat: %s", configs.PublicInstallPageMapFormat)
+	log.Printf("- ClearDirectoryWhenFinished: %s", configs.ClearDirectoryWhenFinished)
 }
 
 func fail(format string, v ...interface{}) {
 	log.Errorf(format, v...)
+	clearDirectoryIfNeeded()
```
richard-bogdan-bitrise: Please do not add this method to failf, as the method name does not indicates the cleanup. Please create a method for example mainE where the cleanup is done and calls failf. See this as example:
https://github.com/bitrise-steplib/bitrise-step-android-build/blob/master/main.go#L328

main.go
```diff
@@ -273,3 +282,31 @@ func validateGoTemplate(publicInstallPageMapFormat string) error {
...
+func clearDirectoryIfNeeded() {
+	configs := createConfigsModelFromEnvs()
+
+	if configs.ClearDirectoryWhenFinished == "true" {
+		log.Printf("Clearing Directory")
+		removeContents(configs.DeployPath)
+	}
+}
+
+func removeContents(pathToDir string) error {
```
richard-bogdan-bitrise: Please add some test cases to test this out.

bitrise.yml
```diff
@@ -51,6 +51,7 @@ workflows:
         - notify_user_groups: $NOTIFY_USER_GROUPS
         - notify_email_list: $NOTIFY_EMAIL_LIST
         - is_enable_public_page: "true"
+        - clear_directory: "true"
```
richard-bogdan-bitrise: Please add test case for single file and directory when this flag is set.

richard-bogdan-bitrise:
@sepperousseau : we are terribly sorry, that you have to wait this much time for getting a review on your PR. I have checked it now, and added my thoughts, please have a look at.
On the other side, I have noticed that you closed the PR. Could you tell me why did you close it?

sepperousseau:
No problem. @richard-bogdan-bitrise 
I closed it because it's been a while and there have been some conflicts since last version.
So I just changed the folder in my setup for every archive so I didn't needed this option anymore.

PR #99, opened 2019-09-04
TOOL-602 Removing curl command

lpusok:
Replaced curl with builtin go http calls.

### DIFF
--- a/uploaders/common.go
+++ b/uploaders/common.go
@@ -110,6 +110,10 @@ func createArtifact(buildURL, token, artifactPth, artifactType string) (string, 
 func uploadArtifact(uploadURL, artifactPth, contentType string) error {
 	log.Printf("uploading artifact")
 
+	netClient := &http.Client{
+		Timeout: time.Second * 3,
+	}
+
 	return retry.Times(3).Wait(5).Try(func(attempt uint) error {
 		file, err := os.Open(artifactPth)
 		if err != nil {
@@ -121,7 +125,7 @@ func uploadArtifact(uploadURL, artifactPth, contentType string) error {
 			}
 		}()
 
-		request, err := http.NewRequest(http.MethodPut, uploadURL, file)
+		request, err := netClient.NewRequest(http.MethodPut, uploadURL, file)
 		if err != nil {
 			return fmt.Errorf("failed to create request, error: %s", err)
 		}
@@ -129,13 +133,13 @@ func uploadArtifact(uploadURL, artifactPth, contentType string) error {
 		request.Header.Add("Content-Type", contentType)
 
 		// Set Content Length manually (https://stackoverflow.com/a/39764726), as it is part of signature in signed URL
-		// fileInfo, err := file.Stat()
-		// if err != nil {
-		// 	return fmt.Errorf("failed to get file info for %s, error: %s", artifactPth, err)
-		// }
-		// request.ContentLength = fileInfo.Size()
+		fileInfo, err := file.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to get file info for %s, error: %s", artifactPth, err)
+		}
+		request.ContentLength = fileInfo.Size()
 
-		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Minute)
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
 		defer cancel()
 		request = request.WithContext(ctx)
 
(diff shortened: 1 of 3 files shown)


uploaders/common_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"testing"
+	"time"
+)
+
+func Test_uploadArtifact(t *testing.T) {
+	const contentType = "image/png"
+
+	file, err := ioutil.TempFile("", "")
+	if err != nil {
+		t.Errorf("setup: failed to create file, error: %s", err)
```
BirmacherAkos: Note
I would use `t.Fatalf` for the initialization part of the test.
E.g: It is unnecessary to run the test cases ff it failed to open/ read the file.

uploaders/common.go
```diff
@@ -110,32 +110,60 @@ func createArtifact(buildURL, token, artifactPth, artifactType string) (string,
...
+			}
+		}()
+
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("failed to read response body, error: %s", err)
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			return fmt.Errorf("invalid status code: %d, headers: %s, body: %s", resp.StatusCode, resp.Header, body)
```
richard-bogdan-bitrise: Note: status code is not invalid, if it is not ok, it indicates that the request failed for some reason. Invalid status code would be 9999 or something that is not commonly used or agreed beforehand.

uploaders/common.go
```diff
@@ -110,32 +110,60 @@ func createArtifact(buildURL, token, artifactPth, artifactType string) (string,
...
+
+		// Set Content Length manually (https://stackoverflow.com/a/39764726), as it is part of signature in signed URL
+		fileInfo, err := file.Stat()
+		if err != nil {
+			return fmt.Errorf("failed to get file info for %s, error: %s", artifactPth, err)
 		}
-		cmd, err := command.NewFromSlice(args)
+		request.ContentLength = fileInfo.Size()
+
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
```
richard-bogdan-bitrise: The client has 10 mins of timeout, but the request could be called 3 times: if the first call timeouts it would not be called more, as the clients timeout already expired.
Why not set timeout to netClient.Timeout/3?

uploaders/common_test.go
```diff
@@ -0,0 +1,138 @@
...
+		var data []byte
+
+		buf := make([]byte, 5)
+		for i := 0; i < 2; i++ {
+			n, err := r.Body.Read(data)
+			if err != nil {
+				t.Errorf("httptest: failed to read body, error: %s", err)
+			}
+			data = append(data, buf[:n]...)
+			time.Sleep(5 * time.Second)
```
richard-bogdan-bitrise: Please consider not using sleep in unit tests, as it greatly slows down testing.

uploaders/common.go
```diff
@@ -110,32 +110,60 @@ func createArtifact(buildURL, token, artifactPth, artifactType string) (string,
...
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to upload artifact, error: %s", err)
 		}
-		cmd.SetStdout(os.Stdout).SetStderr(os.Stderr)
-		return cmd.Run()
+
+		defer func() {
+			if err := resp.Body.Close(); err != nil {
+				log.Errorf("Failed to close response body, error: %s", err)
```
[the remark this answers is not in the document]
godrei: Why? this is not an error we return.

uploaders/common.go
godrei: the client's timeout is per request: https://golang.org/pkg/net/http/#Client

uploaders/common_test.go
godrei: Agree, i removed this test as it is not a different case then the `Happy path` test. I think the original intention was here to test slow server, but it is not enough slow to prove the default timeout 10min

uploaders/common.go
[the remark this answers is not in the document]
richard-bogdan-bitrise: Right, I thought it is a return, but not.

uploaders/common.go
richard-bogdan-bitrise: Right, don't know why I messed it up. Note: it would be still better to use netClient.timeout, as  if you want to change the duration, you have to change it in one place, but this is not that important.

PR #101, opened 2019-09-23
added missing fields

trapacska:
Deploy to Bitrise.io step uploaded an xml in which the corresponding fields were missing. With this fix those fields are passed through also in a way that all the error fields are merged into a single failure field. See the ticket for more technical details.


test/converters/junitxml/junitxml.go
```diff
@@ -25,6 +25,44 @@ func (h *Converter) Detect(files []string) bool {
 	return len(h.files) > 0
 }
 
+// merges Suites->Cases->Error and Suites->Cases->SystemErr field values into Suites->Cases->Failure field
+// with 2 newlines and error category prefix
+// the two newlines applied only if there is a failur emessage already
+func regroupErrors(suites []junit.TestSuite) []junit.TestSuite {
```
richard-bogdan-bitrise: Please add a unit test for this.

test/converters/junitxml/junitxml.go
```diff
@@ -25,6 +25,44 @@ func (h *Converter) Detect(files []string) bool {
 	return len(h.files) > 0
 }
 
+// merges Suites->Cases->Error and Suites->Cases->SystemErr field values into Suites->Cases->Failure field
+// with 2 newlines and error category prefix
+// the two newlines applied only if there is a failure message already
+func regroupErrors(suites []junit.TestSuite) []junit.TestSuite {
+	for testSuiteIndex := range suites {
+		for testCaseIndex := range suites[testSuiteIndex].TestCases {
+			tc := suites[testSuiteIndex].TestCases[testCaseIndex]
```
godrei: NOTE: using these for loops you can spare this line:

```
	for testSuiteIndex, testSuit := range suites {
		for testCaseIndex, tc := range testSuit.TestCases {

		}
	}
```

test/converters/junitxml/junitxml.go
```diff
@@ -34,15 +72,15 @@ func parseTestSuites(filePath string) ([]junit.TestSuite, error) {
 	var testSuites junit.XML
 	testSuitesError := xml.Unmarshal(data, &testSuites)
 	if testSuitesError == nil {
-		return testSuites.TestSuites, nil
+		return regroupErrors(testSuites.TestSuites), nil
 	}
 
 	var testSuite junit.TestSuite
 	if err := xml.Unmarshal(data, &testSuite); err != nil {
 		return nil, errors.Wrap(errors.Wrap(err, string(data)), testSuitesError.Error())
```
godrei: why do you include the `testSuitesError` in the `parseTestSuites ` function's error message?
i feel `parseTestSuites` error should only describe why test suit parsing failed.

test/converters/junitxml/junitxml_test.go
```diff
@@ -23,3 +26,166 @@ func Test_parseTestSuites(t *testing.T) {
 		})
 	}
 }
+
+func Test_regroupErrors(t *testing.T) {
```
godrei: would not it be easier to understand by having xml content test cases?

test/junit/xml.go
```diff
@@ -26,4 +26,13 @@ type TestCase struct {
 	ClassName string   `xml:"classname,attr"`
 	Time      float64  `xml:"time,attr"`
 	Failure   string   `xml:"failure,omitempty"`
+	Error     *Error   `xml:"error,omitempty"`
```
godrei: NOTE: would not it better to have separate models for our junit representation and for the client representations?

test/converters/junitxml/junitxml.go
trapacska: because of the fallback, the first error message might can tell something

Issue #102, opened 2019-09-24
File upload fails when file size is 0


jeremywiebe:
I've been working to resolve an issue where our builds fail if one of the files specified in the Deploy to Bitrise step is 0 bytes.  

The error I'm seeing is:
```
deploy failed, error: failed to upload file artifact, error: non success status code: 501, headers: map[Content-Type:[application/xml] Date:[Tue, 24 Sep 2019 18:43:49 GMT] Server:[AmazonS3] X-Amz-Id-2:[--redacted--] X-Amz-Request-Id:[--redacted--]], body: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NotImplemented</Code><Message>A header you provided implies functionality that is not implemented</Message><Header>Transfer-Encoding</Header><RequestId>--redacted--</RequestId><HostId>--redacted--</HostId></Error>
```

After some research, I believe this is due to the [go http library](https://github.com/golang/go/blob/141b09726dcfc1fa9f04ee9e08e6fa4af00fa57b/src/net/http/request.go#L520) setting transfers to chunked encoding if the file size is 0 _because_ we are explicitly setting the ContentLength header. 

It looks like this was introduced in [release 1.8.0](https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/releases/tag/1.8.0) when the upload was switched to use the go http library from `curl`. 

I have a diff that I think fixes this up here: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/103

jeremywiebe:
Resolved by #103

PR #103, opened 2019-09-24
Fix uploads of zero-byte files

jeremywiebe:
Fixes: Issue #102 

The Go http library sets the TransferEncoding to `"chunked"` if it sees a request that has a ContentLength of zero with a Body set (which is exactly what happens in `uploadArtifact`). 

This PR forces the TransferEncoding to `identity` when the file size is `0` so that this doesn't happen (as documented in the Go library code).


godrei:
Hi @jeremywiebe, thanks for the PR!

PR #105, opened 2019-10-07
Parsing APKs from golang, only fallback to aapt

lpusok:
Added APK parsing using third party apkparser package. This will solve the unreliability of aapt2 binary used so far.


lpusok:
@trapacska pls merge.

PR #113, opened 2019-12-16
zero length file upload

godrei:
Updated `TestXCresult3Converters` as it compared structs with a slice property using `reflect.DeepEqual` it was failing from time to time


bitrise.yml
```diff
@@ -158,6 +160,25 @@ workflows:
...
+    - script:
+        inputs:
+        - content: touch 0_length_file.txt
+    - path::./:
+        title: Zero length file upload test
+        run_if: true
+        inputs:
+        - build_url: $BITRISE_BUILD_URL
+        - build_api_token: $BITRISE_BUILD_API_TOKEN
+        - is_compress: "true"
```
lpusok: Does is_compress affect file_length in any way?

uploaders/common.go
```diff
@@ -134,11 +147,6 @@ func uploadArtifact(uploadURL, artifactPth, contentType string) error {
...
 		}
 
-		// Set Content Length manually (https://stackoverflow.com/a/39764726), as it is part of signature in signed URL
-		fileInfo, err := file.Stat()
-		if err != nil {
-			return fmt.Errorf("failed to get file info for %s, error: %s", artifactPth, err)
-		}
 		request.ContentLength = fileInfo.Size()
 
 		// The go http library will switch the TransferEncoding to 'chunked' if
```
lpusok: Is this part still needed?

bitrise.yml
godrei: pr updated: removed options not affecting this test

uploaders/common.go
godrei: good catch, I am testing it

test/converters/test_utils.go
```diff
@@ -0,0 +1,89 @@
...
+	return reflect.DeepEqual(a, b)
+}
+
+func equivalentTestCases(a, b []junit.TestCase) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	bCopy := append([]junit.TestCase{}, b...)
+	for _, caseA := range a {
```
lpusok: Consider testcases that are in B but they are not in A.

PR #120, opened 2020-03-23
Added support for the "message" attribute of "Failure" for jUnit pars…

lpusok:
…er (used for Test Results Addon)

- Added "Failure" struct, preserving the contents parser ("Value") and added "Message" property which matches the "message" attribure
- The "message" attribute contents is appended to the TestCase Failure contents, similarly as done for the "error" and "system-err" XML subelements.
- Fixed references to the "Failure" field

### DIFF
--- a/test/converters/converters_test.go
+++ b/test/converters/converters_test.go
@@ -57,7 +57,9 @@ func TestXCresult3Converters(t *testing.T) {
 						Name:      "testFailMe",
 						ClassName: "_TtCC17rtgtrghtrgUITests17rtgtrghtrgUITests18rtgtrghtrg3UITests",
 						Time:      0.09049093723297119,
-						Failure:   "file:///Users/tamaspapik/Develop/ios/rtgtrghtrg/rtgtrghtrgUITests/rtgtrghtrgUITests.swift:CharacterRangeLen=0&EndingLineNumber=67&StartingLineNumber=67 - XCTAssertTrue failed",
+						Failure: &junit.Failure{
+							Value: "file:///Users/tamaspapik/Develop/ios/rtgtrghtrg/rtgtrghtrgUITests/rtgtrghtrgUITests.swift:CharacterRangeLen=0&EndingLineNumber=67&StartingLineNumber=67 - XCTAssertTrue failed",
+						},
 					},
 					{
 						Name:      "testLaunchPerformance",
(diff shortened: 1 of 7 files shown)


test/converters/converters_test.go
```diff
@@ -57,7 +57,9 @@ func TestXCresult3Converters(t *testing.T) {
 						Name:      "testFailMe",
 						ClassName: "_TtCC17rtgtrghtrgUITests17rtgtrghtrgUITests18rtgtrghtrg3UITests",
 						Time:      0.09049093723297119,
-						Failure:   "file:///Users/tamaspapik/Develop/ios/rtgtrghtrg/rtgtrghtrgUITests/rtgtrghtrgUITests.swift:CharacterRangeLen=0&EndingLineNumber=67&StartingLineNumber=67 - XCTAssertTrue failed",
+						Failure: &junit.Failure{
+							Value: "file:///Users/tamaspapik/Develop/ios/rtgtrghtrg/rtgtrghtrgUITests/rtgtrghtrgUITests.swift:CharacterRangeLen=0&EndingLineNumber=67&StartingLineNumber=67 - XCTAssertTrue failed",
```
Szabadember: Why is this username hard-coded here?

test/converters/junitxml/junitxml.go
```diff
@@ -22,6 +22,7 @@ func (h *Converter) Detect(files []string) bool {
 			h.files = append(h.files, file)
 		}
 	}
+
```
Szabadember: Is this blank line required?

lpusok: The return statement is usually not cuddled to other lines if there are more than 2 lines. I am using the "wsl" (white space linter) to check these.

test/converters/converters_test.go
lpusok: There is a hardcoded xcresult test result file parsed and checked if matches.

test/converters/junitxml/junitxml_test.go
```diff
@@ -33,152 +33,265 @@ func Test_regroupErrors(t *testing.T) {
 		suites []junit.TestSuite
 		want   []junit.TestSuite
 	}{
-		{"regroup error message", []junit.TestSuite{
-			{TestCases: []junit.TestCase{
-				{Error: &junit.Error{Message: "error message"}},
-			}},
+		{
```
trapacska: pls add a case where message and value  has something

trapacska: or pls prove that it is invalid/forbidden

trapacska: also please add a test where a golden xml contains the field that our implementation currently doesn't support.

NOTE: please next time do the reproduction test case first - to prove the issue is existing and you have a test case that need to be fixed client side.

test/converters/junitxml/junitxml.go
Szabadember: 👍

test/converters/junitxml/junitxml_test.go
lpusok: There is a test case "should append system error, error message, error body to failure" that has both Failure content and message attribute

lpusok: Added test with XML.

PR #124, opened 2020-05-28
added bundletool version input

trapacska:


step.yml
```diff
@@ -196,6 +196,16 @@ inputs:
       value_options:
         - "false"
         - "true"
+  - bundletool_version: "0.13.4"
```
lpusok, line 199: Agreed on adding a fixed version, instead of parsing the possible latest, to prevent unforeseen changes that cause breakage.

bundletool/bundletool.go
```diff
@@ -35,7 +37,7 @@ func New() (Path, error) {
 		return "", err
 	}
 
-	toolPath := filepath.Join(tmpPth, filepath.Base(downloadURL))
+	toolPath := filepath.Join(tmpPth, "bundletool-all.jar")
 
 	return Path(toolPath), ioutil.WriteFile(toolPath, d, 0777)
```
lpusok, line 42: NOTE: Could use io.Copy to write file in a more memory efficient way.

Issue #93, opened 2019-06-28
No way to opt out of universal APK generation


DSteve595:
When deploying a bundle, the step automatically builds a universal APK out of it. However, it's not signed with the same key, so it's not useful to us. We're manually calling `assembleRelease` in addition to `bundleRelease` so that we get an AAB and a universal APK, both signed. It would be good to have an option to prevent the step from generating its own universal APK.

bitce:
Hey @DSteve595,

Thanks for reporting, could you please include a build URL of this happening here?

DSteve595:
@bitce Can I send it privately?

bitce:
Sure thing, just hit us up on our on-site chat 🙂

linean:
We have the same issue as @DSteve595 . In my opinion generating universal APK should be treated as a separate step.

Diego-Franco:
Any update on this one? I have the same issue

radko93:
The same issue, I can't upload my bundle, I don't care about the universal APK.

ahvth-bitrise:
Hi everyone! 👋 Basically we included the universal APK creation in our Deploy to Bitrise.io step for troubleshooting purposes. The step should in no way be changing the actual exported bundle or excluding it from deployment to Bitrise. If for whatever reason, the step is interfering with that functionality, please send us a build URL and we will look into it 😄

RustamG:
> Hi everyone! 👋 Basically we included the universal APK creation in our Deploy to Bitrise.io step for troubleshooting purposes. The step should in no way be changing the actual exported bundle or excluding it from deployment to Bitrise. If for whatever reason, the step is interfering with that functionality, please send us a build URL and we will look into it 😄

@ahvth-bitrise,
The thing is that Install page will contain universal APK instead of AAB that I specified in _Deploy directory or file path_

bitce:
Hi @RustamG!

That's correct. We do have an ongoing issue, but that's that the Deploy step will print this out otherwise (says that an install page is available for the .aab while it's only so for the universal .apk)

This is intended because .aabs are zips and you couldn't directly install them on your device (not due to it being a folder but due it being a distribution method essentially) - that said you can manually allow them on the Apps and Artifacts folder.

arthursady:
@bitce Is there a way to send the email with the .aab public install page automatically? 

I don't want the email to install the apk in a device. I want to be able to download the file without necessarily having access to the bitrise credentials from my account.
[...]

PR #126, opened 2020-11-14
Add option to skip universal APK generation for AABs.

lchrislee:
This is much desired functionality because of the unintended consequences from locally-signed universal APKs. By granting an option (and defaulting it off), we allow the client to determine whether they need this.

Hoping to address #93.

Having issues setting up environments and testing this properly locally since I've never used Go before. Any insight or help with this would be great!


step.yml
```diff
@@ -223,6 +223,13 @@ inputs:
         The token required to authenticate with the API.
       is_dont_change_value: true
       is_sensitive: true
+  - generate_universal_apk_if_none: "false"
+    opts:
+      title: "Force generate universal APK for AABs"
```
mateherber, line 228: Hi @lchrislee ! Thank you for contribution. Please make the default value of this input true so it will not be a breaking change on the step's contract.

main.go
```diff
@@ -39,6 +39,7 @@ type Config struct {
 	AppSlug                       string `env:"BITRISE_APP_SLUG,required"`
 	AddonAPIBaseURL               string `env:"addon_api_base_url,required"`
 	AddonAPIToken                 string `env:"addon_api_token"`
+  GenerateUniversalApkIfNone    bool   `env:"generate_universal_apk_if_none,opt[true,false]"`
```
mateherber, line 42: I think my merge messed with the formatting, I'd be glad if you could fix that

step.yml
lchrislee: Good catch! Thanks for looking into this.

On a side note, I have no idea how unit tests passed when I defaulted to false 🤷. I'd love to update the unit tests but I haven't had time to look into my setup since opening the PR.

mateherber: Hi! Unfortunately not all of our steps are covered fully with tests, we're continuously looking into improving our steps and trying to prioritize work based on their usage and importance. We're planning on tackling the real issue defined in #93 so this flag is just a temporary measure. You can just go with setting the flag true. We'll open an other issue to clean the apk generation.

mateherber (approved):

PR #127, opened 2020-12-11
[STEP-244] Permanent Download URLs

mateherber:
Add permanent download url for artifacts to the output

Introduced a new output `BITRISE_PERMANENT_DOWNLOAD_URL_MAP` to hold
permanent urls for the artifacts uploaded by the step. Introduced a
new input `permanent_download_url_map_format` to control the format
of the output.

We received a request that it would be helpful if this step made
permanent download links available on its output for other steps to use.
Engagement team is responsible for generating these links. [EN-492]

[STEP-244]

[EN-492]: https://bitrise.atlassian.net/browse/EN-492
[STEP-244]: https://bitrise.atlassian.net/browse/STEP-244


main.go
```diff
@@ -307,42 +333,52 @@ func deploy(clearedFilesToDeploy []string, config Config) (map[string]string, er
...
 	}
-	return publicInstallPages, nil
+	return artifactURLCollection, nil
+}
+
+func fillURLMaps(artifactURLCollection ArtifactURLCollection, artifactURLs uploaders.ArtifactURLs, apk string, isPublic bool) {
+	if artifactURLs.PublicInstallPageURL != "" {
+		artifactURLCollection.PublicInstallPageURLs[filepath.Base(apk)] = artifactURLs.PublicInstallPageURL
+	} else if isPublic {
+		log.Warnf("is_enable_public_page is set, but public download isn't allowed for this type of file")
```
godrei, line 377: I am not sure if moving this warning here makes sense, previously we printed this only in case of fileupload, now we handle it for each extension.

I think the previous version was not fine either, as the user did not misconfigure anything, but we still print a warning. IMO we should explain for what file the step will generate a public install page and remove this warning message.

godrei: previously we filled the `publicInstallPages` only in case of certain file types, now we fill it based on if the file has got the URL or not. Did not we change here what gets into the `publicInstallPages `? Like a `.aab` will not get a public install page?

step.yml
```diff
@@ -226,3 +234,15 @@ outputs:
 
         - $BITRISE_DEPLOY_DIR/ios_app.ipa=>https://ios_app/public/install/page
         - $BITRISE_DEPLOY_DIR/android_app.apk=>https://android_app/public/install/page|$BITRISE_DEPLOY_DIR/ios_app.ipa=>https://ios_app/public/install/page
+  - BITRISE_PERMANENT_DOWNLOAD_URL_MAP:
```
godrei, line 237: Mig it would wort to update `BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP` description too with `public_install_page_url_map_format`

godrei (commented):
Pls check my notes.

main.go
mateherber: I wanted to change the text of the warning, thanks :)

mateherber: Yes you're right I'm gonna add an additional flag

step.yml
mateherber: 👍

godrei (dismissed):

step.yml
```diff
@@ -220,9 +228,21 @@ outputs:
...
         Examples:
 
         - $BITRISE_DEPLOY_DIR/ios_app.ipa=>https://ios_app/public/install/page
         - $BITRISE_DEPLOY_DIR/android_app.apk=>https://android_app/public/install/page|$BITRISE_DEPLOY_DIR/ios_app.ipa=>https://ios_app/public/install/page
+  - BITRISE_PERMANENT_DOWNLOAD_URL_MAP:
+    opts:
+      title: Map of filenames and Permanent Download URLs
+      description: |-
+        Permanent Download URLs by the artifact's file path.
+
```
BanyikAnna, line 242: If there are multiple artifact URLs, you can customize their format.

step.yml
```diff
@@ -220,9 +228,21 @@ outputs:
...
 
         Examples:
 
         - $BITRISE_DEPLOY_DIR/ios_app.ipa=>https://ios_app/public/install/page
         - $BITRISE_DEPLOY_DIR/android_app.apk=>https://android_app/public/install/page|$BITRISE_DEPLOY_DIR/ios_app.ipa=>https://ios_app/public/install/page
+  - BITRISE_PERMANENT_DOWNLOAD_URL_MAP:
+    opts:
+      title: Map of filenames and Permanent Download URLs
+      description: |-
+        Permanent Download URLs by the artifact's file path.
```
BanyikAnna, line 241: Maybe: "Permanent Download URL generated by the artifacts' file paths. " would that describe the situ well?

step.yml
```diff
@@ -136,7 +136,15 @@ inputs:
       description: |-
         Specify the output pattern with golang template: [https://golang.org/pkg/text/template](https://golang.org/pkg/text/template)
       is_required: true
-      is_expand: false                    
+      is_expand: false
+  - permanent_download_url_map_format: "{{range $index, $element := .}}{{if $index}}|{{end}}{{$element.File}}=>{{$element.URL}}{{end}}"
+    opts:
+      title: Format for the BITRISE_PERMANENT_DOWNLOAD_URL_MAP output
+      summary: You can change the format of the BITRISE_PERMANENT_DOWNLOAD_URL_MAP output
```
BanyikAnna, line 143: I'd rephrase it to something like this: You can customize the format of the `BITRISE_PERMANENT_DOWNLOAD_URL_MAP` output so that it fits the next Step you wish to use the output with.

step.yml
```diff
@@ -136,7 +136,15 @@ inputs:
...
+      description: |-
+        Specify the output pattern with golang template: [https://golang.org/pkg/text/template](https://golang.org/pkg/text/template)
```
BanyikAnna, line 145: Customize the output pattern with this golang template: [https://golang.org/pkg/text/template](https://golang.org/pkg/text/template)

step.yml
```diff
@@ -136,7 +136,15 @@ inputs:
       description: |-
         Specify the output pattern with golang template: [https://golang.org/pkg/text/template](https://golang.org/pkg/text/template)
```
BanyikAnna, line 137: Customize the output pattern with this golang template: [https://golang.org/pkg/text/template](https://golang.org/pkg/text/template)

step.yml
mateherber: This after line 241 or instead something?

step.yml
mateherber: These are rather "File Path - URL" pairs for the respective artifact.

step.yml
mateherber: The referenced link is rather a template language, not a template itself. Using it you can create templates.

BanyikAnna: Customize the output pattern by using this golang template language: https://golang.org/pkg/text/template

step.yml
BanyikAnna: "If there are multiple artifact URLs, you can customize their format." should come before the example you provided below. It can be a continuation of line 242 or right after it, but I'd definitely add it before the examples and the default format. :)

step.yml
BanyikAnna: Then something like this: This input creates a permanent Download URL for artifacts. The URL can be shared in any communication channel and it will not expire.

main.go
```diff
@@ -90,55 +97,74 @@ func main() {
...
-				File: file,
-				URL:  url,
-			})
-		}
+func exportInstallPages(artifactURLCollection ArtifactURLCollection, config Config) error {
+	if len(artifactURLCollection.PublicInstallPageURLs) > 0 {
+		pages := mapURLsToInstallPages(artifactURLCollection.PublicInstallPageURLs)
 
 		if err := tools.ExportEnvironmentWithEnvman("BITRISE_PUBLIC_INSTALL_PAGE_URL", pages[0].URL); err != nil {
 			return fmt.Errorf("failed to export BITRISE_PUBLIC_INSTALL_PAGE_URL, error: %s", err)
```
adborbas, line 119: `failed to do magic, error: no magic` is not a recommended pattern. Rather use `failed to do magic: no magic`. 
Like in https://godoc.org/github.com/pkg/errors#Wrap
```suggestion
			return fmt.Errorf("failed to export BITRISE_PUBLIC_INSTALL_PAGE_URL: %s", err)
```

Please apply below as well.

mateherber: Not fair! Hasn't even touched that :P

adborbas (approved):

PR #128, opened 2020-12-16
step description updates

BanyikAnna:
Updated the Step description and the input fields as well.


step.yml
```diff
@@ -2,31 +2,52 @@ title: "Deploy to Bitrise.io - Apps, Logs, Artifacts"
...
+  3. Set the **Notify: Emails** sensitive input. It sends the public install URL in an email to the email addresses provided here. If you’re adding multiple email address, make sure to separate them with commas. 
+  The recipients do not have to be in your Bitrise team. Please note that if the email address is associated with a Bitrise account, the user must be [watching](https://devcenter.bitrise.io/builds/configuring-notifications/#watching-an-app) the app.
+  4. The **Enable public page for the App?** required input is set to `true` by default. It creates a long and random URL which can be shared with those who do not have a Bitrise account. 
+  If you set this input to `false`, the **Notify: Emails** input will be ignored and the **Notify: User Roles** will receive the build URL instead of the public install URL.
+  5. With the **Compress the artifacts into one file?** required input set to `true`, you can compress the artifacts found in the Deploy directory into a single file.
+  You can specify a custom name for the zip file with the `zip_name` option. Otherwise the default Deploy directory will be used.
```
mateherber, line 20: `Otherwise the default Deploy directory will be used`

Maybe somthing like 

`Otherwise the default Deploy directory's name will be used`

would be better?

step.yml
```diff
@@ -2,31 +2,52 @@ title: "Deploy to Bitrise.io - Apps, Logs, Artifacts"
...
+  You can specify a custom name for the zip file with the `zip_name` option. Otherwise the default Deploy directory will be used. 
+  If the **Compress the artifacts into one file?** is set to `false`, the artifacts in the Deploy directory will be deployed separately.
+  6. With the **Format for the BITRISE_PUBLIC_INSTALL_PAGE_URL_MAP output** required input field, you can customize the output format of the public install page’s multiple artifact URLs so that the next Step can render the output (for example, our **Send a Slack message** Step). 
+  Provide a language template description using [https://golang.org/pkg/text/template]([https://golang.org/pkg/text/template) so that the **Deploy to Bitrise.io** Step can build the required custom output.
+  7. With the **Format for the BITRISE_PERMANENT_DOWNLOAD_URL_MAP output** required input, you can customize the output format of the `BITRISE_PERMANENT_DOWNLOAD_URL_MAP` so that the next Step can render the output. 
+  This input creates a permanent URL containing all the artifacts the build has generated.
```
mateherber, line 25: `This input creates a permanent URL containing all the artifacts the build has generated.`

The input will be used to generate the related output in the specified format. The output contains multiple permanent urls for the multiple artifacts.

Can't we reuse the same content that is written for the public format?

step.yml
```diff
@@ -2,31 +2,52 @@ title: "Deploy to Bitrise.io - Apps, Logs, Artifacts"
...
+  Provide a language template description using [https://golang.org/pkg/text/template]([https://golang.org/pkg/text/template) so that the **Deploy to Bitrise.io** Step can build the required custom output.
+  8. The **Test API's base URL** and the **API Token** input fields are automatically populated for you.
```
mateherber, line 27: Maybe this shouldn't even be mentioned here as we give no context what they are for.

step.yml
```diff
@@ -2,31 +2,52 @@ title: "Deploy to Bitrise.io - Apps, Logs, Artifacts"
...
+  
+  #### Configuring the Debug section of the Step
+  
+  If you wish to use any of the Step’s debug features, set the following inputs:
+  1. In the **Name of the compressed artifact (without .zip extension)** input you can add a custom name for the compressed artifact. If you leave this input empty, the default `Deploy directory` name is used.
+  Please note that this input only works if you set the **Compress the artifacts into one file?** input to `true`.
+  2. The **Bitrise Build URL** and the **Bitrise Build API Token** inputs are automatically populated.
```
mateherber, line 34: Maybe this can be left out too?

step.yml
```diff
@@ -2,31 +2,52 @@ title: "Deploy to Bitrise.io - Apps, Logs, Artifacts"
...
+  3. If **The Enable Debug Mode** required input is set to `true`, the Step prints more verbose logs. It is `false` by default. 
+  4. If you wish to generate an Android App Bundle and an APK which you can test, set a specific [bundletool version]((https://github.com/google/bundletool/releases) in the **Bundletool version** required input.
```
mateherber, line 36: The bundletool is used for generating the APK from the AAB. For the aab generation it is not needed. This input is required so even if you don't set it we will use the default bundletool when we see an AAB.

I think currently the description implies that if you need AAB you have to change the value here somehow.

step.yml
```diff
@@ -2,31 +2,52 @@ title: "Deploy to Bitrise.io - Apps, Logs, Artifacts"
...
 
-  1. Set the value for the Deploy directory of file path. The default is the $BITRISE_DEPLOY_DIR environment variable which is exposed by the Bitrise CLI.
-  2. Set the **Notify: User Roles** and the **Notify: Emails** inputs to configure notifications for the users. Note that the **Notify: Emails** input is only used if a public install page is created and you must use a Secret to set it.
-  3. Optionally, you can create a zip file for all your build artifacts by setting the **Compress the artifacts into one file?** input to `true`. 
-
-  Get a public install page that allows testers to install the app on their devices:
-
-  1. Set the **Enable public page for the App?** input to `true`.
-  2. Set the **Compress the artifacts into one file?** input to `false`. If all the contents of the deploy directory are compressed into a single zip file, there won't be a public install page.
+  1. Set the value for the **Deploy directory or file path** required input. It can be a path to a directory or a file. If you provide a directory, everything in that directory gets uploaded recursively.
```
mateherber, line 11: `If you provide a directory, everything in that directory gets uploaded recursively.`

Sorry I gave you misinformation. I've double checked and if the directory is provided only files in the directory **excluding** the subdirectories will be uploaded.

Everything will be uploaded only if compression is enabled.

This is written well in the description of the input.

step.yml
```diff
@@ -50,11 +71,15 @@ inputs:
       title: "Deploy directory or file path"
       description: |
         Specify the directory or file path which will be deployed.
+        
         If the specified path is a directory, then every file
-        in the specified directory will be deployed, sub-directories, however,
-        won't be uploaded. To upload the directory's content
-        recursively, you should use the *Compress* option
-        which will compress the whole directory, with
+        in the specified directory will be deployed. If, however,
```
mateherber, line 76: `If, however, the specified path is a sub-directory, it won't be uploaded. `

The specified path will always be a directory (never a subdirectory because it is relative). Every file in the specified directory will be uploaded and **no** files in the subdirectories found in the specified directory

step.yml
```diff
@@ -132,29 +158,41 @@ inputs:
...
   - permanent_download_url_map_format: "{{range $index, $element := .}}{{if $index}}|{{end}}{{$element.File}}=>{{$element.URL}}{{end}}"
     opts:
       title: Format for the BITRISE_PERMANENT_DOWNLOAD_URL_MAP output
-      summary: You can customize the format of the BITRISE_PERMANENT_DOWNLOAD_URL_MAP output so that it fits the next Step you wish to use the output with.
+      summary: You can customize the format of the `BITRISE_PERMANENT_DOWNLOAD_URL_MAP` output so that the next Step can render the output.
       description: |-
-        Customize the output pattern with this golang template: [https://golang.org/pkg/text/template](https://golang.org/pkg/text/template)
+        Provide a language template description using [https://golang.org/pkg/text/template]([https://golang.org/pkg/text/template)
+        so that the **Deploy to Bitrise.io** Step can build the required custom output for the permanent download URL. 
+        This URL can be shared in any communication channel and it will not expire.
```
mateherber, line 174: This line and next ones should go to the output. Also the output contains not a single URL but a set of urls for the different uploaded artifacts

step.yml
```diff
@@ -208,11 +246,11 @@ inputs:
     opts: 
       category: Debug
       title: "Bundletool version"
-      summary: The version of bundletool you want to use.
+      summary: The specific bundletool version you want to use.
       description: |-
-        The version you specify will be installed.
-
-        Versions are available here: https://github.com/google/bundletool/releases
+        If you wish to generate an Android App Bundle and an APK which you can test,
```
mateherber, line 251: The same applies here as in the description.

step.yml
```diff
@@ -238,7 +276,7 @@ outputs:
     opts:
       title: Map of filenames and Permanent Download URLs
       description: |-
-        This input creates a permanent Download URL for artifacts. The URL can be shared in any communication channel and it will not expire.
+        This input creates a permanent Download URL for all artifacts. The URL can be shared in any communication channel and it will not expire.
```
mateherber, line 279: This is an output of the step so the `This input` part might be confusing. maybe `This output contains a permanent...` would be more clear

mateherber (changes requested):

step.yml
BanyikAnna: Thanks @mateherber ! :). I'll correct it ;)

step.yml
BanyikAnna: We can,  but then I freaked out since I did not mention the permanent side of things which is quite important with this input so I added that part. I like this solution: The input will be used to generate the related output in the specified format. The output contains multiple permanent urls for the multiple artifacts.

step.yml
BanyikAnna: Yeah I get it but I wanted to let the users know that they don't have to touch this since these are there for them automatically..

step.yml
```diff
@@ -8,21 +8,24 @@ description: |-
 
   ### Configuring the Step
 
-  1. Set the value for the **Deploy directory or file path** required input. It can be a path to a directory or a file. If you provide a directory, everything in that directory gets uploaded recursively. 
-  If you provide only a file, then only that file gets uploaded. The default value is the `$BITRISE_DEPLOY_DIR` Env Var which is exposed by the Bitrise CLI.
+  1. Set the value for the **Deploy directory or file path** required input. The default value is the `$BITRISE_DEPLOY_DIR` Env Var which is exposed by the Bitrise CLI.
+  It can be a path to a directory or a file. 
+  If you provide a directory, everything in that directory, excluding sub-directories, gets uploaded recursively.
```
mateherber, line 13: Everything in the directory gets uploaded and nothing deeper. so recursively is not needed.

step.yml
```diff
@@ -8,21 +8,24 @@ description: |-
...
   The recipients do not have to be in your Bitrise team. Please note that if the email address is associated with a Bitrise account, the user must be [watching](https://devcenter.bitrise.io/builds/configuring-notifications/#watching-an-app) the app.
   4. The **Enable public page for the App?** required input is set to `true` by default. It creates a long and random URL which can be shared with those who do not have a Bitrise account. 
   If you set this input to `false`, the **Notify: Emails** input will be ignored and the **Notify: User Roles** will receive the build URL instead of the public install URL.
   5. With the **Compress the artifacts into one file?** required input set to `true`, you can compress the artifacts found in the Deploy directory into a single file.
-  You can specify a custom name for the zip file with the `zip_name` option. Otherwise the default Deploy directory will be used. 
+  You can specify a custom name for the zip file with the `zip_name` option. If you don't specify one, the default Deploy directory will be used.
```
mateherber, line 23: it is still confusing to me. The deploy directory will be used for? (its name will be used for naming the zip file)

step.yml
```diff
@@ -33,7 +36,8 @@ description: |-
   Please note that this input only works if you set the **Compress the artifacts into one file?** input to `true`.
   2. The **Bitrise Build URL** and the **Bitrise Build API Token** inputs are automatically populated.
   3. If **The Enable Debug Mode** required input is set to `true`, the Step prints more verbose logs. It is `false` by default. 
-  4. If you wish to generate an Android App Bundle and an APK which you can test, set a specific [bundletool version]((https://github.com/google/bundletool/releases) in the **Bundletool version** required input.  
+  4. If you need a specific [bundletool version]((https://github.com/google/bundletool/releases) other than the default value, you can modify the value of the **Bundletool version** required input. 
+  This input generates an APK from an Android App Bundle so that you can test the APK.
```
mateherber, line 40: not the input generates the apk but the bundletool

step.yml
```diff
@@ -248,9 +240,9 @@ inputs:
       title: "Bundletool version"
       summary: The specific bundletool version you want to use.
       description: |-
-        If you wish to generate an Android App Bundle and an APK which you can test,
-        set a specific [bundletool version]((https://github.com/google/bundletool/releases) 
-        in the **Bundletool version** required input. 
+        If you need a specific [bundletool version]((https://github.com/google/bundletool/releases) other than the default version,
+        you can modify the value of the **Bundletool version** required input. 
+        This input generates an APK from an Android App Bundle so that you can test the APK.
```
mateherber, line 245: see comment above

step.yml
```diff
@@ -276,11 +268,10 @@ outputs:
     opts:
       title: Map of filenames and Permanent Download URLs
       description: |-
-        This input creates a permanent Download URL for all artifacts. The URL can be shared in any communication channel and it will not expire.
-
-        If there are multiple artifact URLs, you can customize their format.
+        The output contains multiple permanent Download URL for all artifacts. The URLs can be shared in any communication channel and they won't expire.
```
mateherber, line 271: it contains 1 url for each artifact

mateherber (approved):

PR #132, opened 2021-02-10
Remove universal APK generation

istvankovacs-bitrise:
### Checklist

- [X] I've read and accepted the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)
- Requires _MINOR_ [version update](https://semver.org/)

### Context
Since the _Export Universal APK_ Step is capable of generating a universal APK from a certain AAB, and furthermore, it can use the same certificate for signing as for the original artifact(s), it will become the recommended way. Due to that, this capability is not required in this Step anymore.

### Changes
- Universal APK generation and debug signing parts removed.
- Related tests removed.

### Investigation details
Please see [this](https://bitrise.atlassian.net/wiki/spaces/STEP/pages/1168048427/Investigate+universal+APK+exporting) Confluence document for the details of the investigation.


mateherber:
Removing the flag from the input would be a major version change right? Would we want to remove without deprecation window or first we should release a version where we deprecate and then remove. WDYT @adborbas / @godrei ?

istvankovacs-bitrise:
I wasn't sure about the version, I wanted to bring this up later by the time I'm releasing it.

The deprecation is a good idea. Do we have a source on how to deprecate an input?

lpusok:
We could add a warning if aabs are detected but no universal apks generated from them. In addition we could have a transition period where we set the input to false and warn anyone trying to use it (not sure about this one)

istvankovacs-bitrise:
We will wait for the removal of the feature until the next major release.

PR #150, opened 2022-01-17
[BINS-246] Add skipped field parsing to the xc converter

trapacska:
based ont this ticket: https://bitrise.atlassian.net/browse/BINS-246

For both iOS and Android there are a possibility of skipping specific test cases when testing. Android has a build in junit4 generator thus working fine and sends us the skipped statuses for the test cases.

However for iOS this step converts xcresult file into junit4 but for some reason the skipped status for test cases were completely skipped/forgotten.

This PR is about to convert the skipped status as well.

This should be a patch version bump.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -20,23 +20,10 @@ workflows:
   ci:
     before_run:
     - audit-this-step
+    - test
     after_run:
     - _zero_length_file_upload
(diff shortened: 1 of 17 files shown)


ofalvai (approved):
LGTM!

PR #151, opened 2022-01-24
[BINS-246] Fixing missing skipped test in XCResult3

viktorbenei:
### Checklist

### Version

### Context

xcresult3 converter's output includes the `skipped` tests as `successful` - bug fix

### Changes

- fix of skipped handling
- added unit test for xcresult3's `XML()` func


test/converters/xcresult3/converter_test.go
```diff
@@ -0,0 +1,74 @@
...
+		filepath.Join("../../testdata/", pathInTestdataDir),
+		dirPathToCopyInto,
+		false,
+	)
+	require.NoError(t, err)
+	return filepath.Join(dirPathToCopyInto, pathInTestdataDir)
+}
+
+func TestConverter_XML(t *testing.T) {
+	t.Run("xcresults3 success-failed-skipped-tests.xcresult", func(t *testing.T) {
```
godrei, line 30: NOTE: what is the reason for running the single test case as a subtest?

viktorbenei:
> is there a way to add sample files to a different repo to prevent bloating of this one? Currently this one is used: https://github.com/bitrise-io/sample-artifacts/tree/master/xcresults

xcresult3 sample moved into the sample-artifacts repo

can you please approva that commit first, and once merged the tests should pass here: https://github.com/bitrise-io/sample-artifacts/pull/20

godrei:
Within the current circumstances, I'm fine with the unit test for `xcresult3` converter.
But in long term, this could lead to issues (like the CI can be run only on macOS). Long term it would be better to mock the actual `xcrun` command execution.

BirmacherAkos:
@viktorbenei Will this solve the following ticket?
https://bitrise.atlassian.net/browse/STEP-1740

viktorbenei:
> @viktorbenei Will this solve the following ticket? https://bitrise.atlassian.net/browse/STEP-1740

Yep @BirmacherAkos I think that ticket is the same as https://bitrise.atlassian.net/browse/BINS-246

test/converters/xcresult3/converter_test.go
viktorbenei: to be able to `defer` the temp directory cleanup

that said, as we only have a single test in it right now it doesn't really matter, but if we'd have another one this way we could generate and cleanup separate temp dirs for them

godrei (approved):

PR #152, opened 2022-02-16
[BINS-319] junit report generation fix when report includes a flaky rerun

viktorbenei:
### Checklist

### Version

I'd suggest a *MINOR* [version update](https://semver.org/).

There should be no breaking changes, but the logic of how we decide for xcresult3 report if a given test is failed changed.
Previously it was based on whether the code could find a failure for it in the failure summaries, while the new code now depends on the test's `TestStatus.Value == "Failure"`.

That said, there's one semi-breaking-change: the Failure Message in the xcresult3->junit generated report changed. Previously it included a url encoded reference (which is what the "issues" summary has as an error message), while the new solution now constructs the error message itself with the format: `file-path:line-number - error message`. You can see an example of this change here: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/152/files#diff-8f7805bc321215ea75557289f345cd1e7f9d2a325d04030fe5ed93bb93dde224L61

**As far as we know this isn't a breaking change, but might worth a chat.**

### Context

Fixing the issue when the xcresult3 report includes a flaky test,
performed in "retry on failure" mode,
where the exact same test case first fails,
and then succeeds on the retry/rerun.

The previous implementation generated a junit report
which included both runs of the test case,
but both were marked as failed,
while the expected output would be a junit
report with a failed and then a successful run for the same (flaky) test case.

### Changes

The primary change is that xcresult3 `XML()` no longer determines whether the test failed or not by checking the `test-suite-name` + `test-case-name` in the "issues" summary list, because if you have "retry on failure" enabled then the report can include the same test case (where `test-suite-name` + `test-case-name` is the same) multiple times.

The previous logic simply picked the first match, so when the report included the same test case once as failed and once as successful (success on retry) the previous logic falsly listed the test case as failed 2 times. The new code will now list it as 1 failed 1 success in the generated junit report.

### Investigation details

x

### Decisions

x

### DIFF
--- a/test/converters/xcresult/testsummariesplist.go
+++ b/test/converters/xcresult/testsummariesplist.go
@@ -24,7 +24,8 @@ func collapseSubtestTree(data Subtests) (tests Subtests) {
 }
 
 // Tests returns the collapsed tree of tests
-func (summaryPlist TestSummaryPlist) Tests() map[string]Subtests {
+func (summaryPlist TestSummaryPlist) Tests() ([]string, map[string]Subtests) {
+	keyOrder := []string{}
 	tests := map[string]Subtests{}
 	var subTests Subtests
 	for _, testableSummary := range summaryPlist.TestableSummaries {
@@ -35,9 +36,12 @@ func (summaryPlist TestSummaryPlist) Tests() map[string]Subtests {
 	for _, test := range subTests {
 		// TestIdentifier is in a format of testID/testCase
 		testID := strings.Split(test.TestIdentifier, "/")[0]
+		if _, found := tests[testID]; !found {
+			keyOrder = append(keyOrder, testID)
+		}
 		tests[testID] = append(tests[testID], test)
 	}
-	return tests
+	return keyOrder, tests
 }
 
 // Test ...
--- a/test/converters/xcresult/testsummariesplist.go
+++ b/test/converters/xcresult/testsummariesplist.go
@@ -25,7 +25,7 @@ func collapseSubtestTree(data Subtests) (tests Subtests) {
 
 // Tests returns the collapsed tree of tests
 func (summaryPlist TestSummaryPlist) Tests() ([]string, map[string]Subtests) {
-	keyOrder := []string{}
+	var keyOrder []string
 	tests := map[string]Subtests{}
 	var subTests Subtests
 	for _, testableSummary := range summaryPlist.TestableSummaries {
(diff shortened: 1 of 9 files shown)


test/converters/xcresult/testsummariesplist.go
```diff
@@ -24,7 +24,8 @@ func collapseSubtestTree(data Subtests) (tests Subtests) {
 }
 
 // Tests returns the collapsed tree of tests
-func (summaryPlist TestSummaryPlist) Tests() map[string]Subtests {
+func (summaryPlist TestSummaryPlist) Tests() ([]string, map[string]Subtests) {
+	keyOrder := []string{}
```
godrei, line 28: NOTE: better to declare a slice using `var keyOrder []string`: https://gist.github.com/adamveld12/c0d9f0d5f0e1fba1e551#declaring-empty-slices

test/converters/xcresult3/action_test_plan_summary.go
```diff
@@ -39,9 +39,10 @@ type Name struct {
 }
 
 // tests returns ActionTestSummaryGroup mapped by the container TestableSummary name.
-func (s ActionTestPlanRunSummaries) tests() map[string][]ActionTestSummaryGroup {
+func (s ActionTestPlanRunSummaries) tests() ([]string, map[string][]ActionTestSummaryGroup) {
 	summaryGroupsByName := map[string][]ActionTestSummaryGroup{}
 
+	testSuiteOrdering := []string{}
```
godrei, line 45: NOTE: based on this: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/152/files#diff-f1f03b66d8394133fe7409796d0c4dd9628abd604720db68b01bd3dbe8be15e1R28

`testSuiteOrdering` could be `testSuiteOrder` or the other way around

test/converters/xcresult3/action_test_plan_summary.go
```diff
@@ -53,14 +54,15 @@ func (s ActionTestPlanRunSummaries) tests() map[string][]ActionTestSummaryGroup
 			}
 
 			summaryGroupsByName[name] = tests
+			testSuiteOrdering = append(testSuiteOrdering, name)
```
godrei, line 57: `summaryGroupsByName[name] = tests` will keep the last tests with the same name,
while `testSuiteOrdering = append(testSuiteOrdering, name)` will store all the names (even the same name multiple times), shouldn't we use:

```
			if _, found := summaryGroupsByName[name]; !found {
				testSuiteOrdering = append(testSuiteOrdering, name)
			}
			summaryGroupsByName[name] = tests
```

godrei (approved):

PR #156, opened 2022-05-25
[BINS-515] xcresult to junit conversion optimization

viktorbenei:
This PR affects only the xcresult3 to junit conversion.

In that case it'll run all the test case conversions in parallel.

## Why?

The `xcresulttoolGet` call, which we run for every test case to determine if it has an attached screenshot, takes about 20ms. Ref: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/fd92ed83c3928d8b93d39dcad00633f2df0b1bfb/test/converters/xcresult3/action_test_summary_group.go#L97

This means that for a test report which has 4000 test cases these xcresulttool calls will take about 80 seconds on an M1 Mac machine.

This PR improves this a bit by doing these calls in parallel.

## Data

On my personal M1 Mac with 10 cores this change improved the conversion time from 80 seconds to ~20 seconds.

On a Standard Bitrise macOS VM it improved from ~250 seconds to ~150 seconds.

## Further improvements

If we could get rid of that `xcresulttoolGet` call at https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/fd92ed83c3928d8b93d39dcad00633f2df0b1bfb/test/converters/xcresult3/action_test_summary_group.go#L97
we could get the conversion time down from ~250 seconds to ~12 seconds (!!) on a Standard Bitrise VM (tested, with a 4000 test case report which had no screenshots, so I just removed the `exportScreenshots` call).

Unfortunately I couldn't find a way yet to determine whether a test case has a related screenshots without that call, that needs further investigation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+# TODO
+
+- Add timestamps to detection, conversion, upload - so we can tell from the logs where the time is spent the most
+
 # Bitrise Deploy Step
 
 Generic Bitrise deployer Step.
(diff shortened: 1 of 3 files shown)


viktorbenei (commented):
lgtm 👍  (Laci's changes)

PR #161, opened 2022-10-10
CI-35

vshah23:
### Checklist

### Version

### Context

This step was failing on linux machines when producing archives for directories for pipeline intermediates as `zip` command is not available.

https://app.bitrise.io/app/67f5caf5b7670167/pipelines/2ae871be-db66-4364-ad73-94fd8c55c889

### Changes

  - Update dependencies
  - Pipeline intermediate directories are now archived using `tar` command instead of `zip`.

### Decisions

Decided not to compress the archive as the gains in this regard are negligible and it would require a subsequent change in the pull-intermediate-files step.

On versioning, from @godrei: 

> I think it should be a patch: https://semver.org/#spec-item-6
> Because this fixes an incorrect behaviour and transferring dirs between stages as zip is not a public api/functionality, it is just a private implementation detail.
> If we change the archive format for deployed artifacts, that would be a major change, because those artifacts supposed to be used by users directly,


vshah23:
E2E test here: https://github.com/bitrise-steplib/bitrise-step-pull-intermediate-files/pull/2

main.go
```diff
@@ -296,6 +297,23 @@ func collectFilesToDeploy(absDeployPth string, config Config, tmpDir string) (fi
...
 }
 
+func compressAsTar(sourceDirPath, destinationTarPath string, isContentOnly bool) error {
+	fmt.Println()
+	log.Infof("Compressing directory...")
+
+	// -c - create a new archive
+	// -f - the next argument is the name of the output archive
+	// Note: recursive behavior is default in tar
+	cmd := command.New("/usr/bin/tar", "cf", destinationTarPath, sourceDirPath)
```
ofalvai, line 307: I don't think it's a good idea to use an absolute path here, we can't make sure it's the same on all systems. For example, on M1 macOS, `/usr/bin` is read-only, so any manually installed `tar` will live in a different directory (such as `/opt/homebrew/bin`).

main.go
```diff
@@ -296,6 +297,23 @@ func collectFilesToDeploy(absDeployPth string, config Config, tmpDir string) (fi
 	return filesToDeploy, nil
 }
 
+func compressAsTar(sourceDirPath, destinationTarPath string, isContentOnly bool) error {
```
ofalvai, line 300: If we are not compressing the tar, isn't it a bit misleading to call this `compress` everywhere? I think the correct term for this whole thing is to create an archive (BTW tar is Tape ARchive 😉 )

deployment/collector.go
```diff
@@ -195,13 +195,13 @@ func (c Collector) zipDirectories(items []DeployableItem) ([]DeployableItem, err
 	return items, nil
 }
 
-func (c Collector) zipDir(path string) (string, error) {
+func (c Collector) compressDir(path string) (string, error) {
 	name := filepath.Base(path)
-	targetPth := filepath.Join(c.temporaryFolder, name+".zip")
+	targetPath := filepath.Join(c.temporaryFolder, name+".zip")
```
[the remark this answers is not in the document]
vshah23: `tar` doesn't seem to care what the extension is. I considered leaving this alone as I thought the method might be used in multiple places, but upon checking that it's not true. Fixing it.

ofalvai (dismissed):

godrei (approved):

StasDoskalenkoHover:
I think now it preserves the whole folder structure instead of just the end directory, which messes things up on OSX.

treastrain:
> I think now it preserves the whole folder structure instead of just the end directory, which messes things up on OSX.

Same.
I'm using this in our macOS instance, but have had to make some disruptive modifications to my workflow/pipeline.

PR #168, opened 2023-01-18
Duplicated artifacts

godrei:
### Checklist

### Version

### Context

This PR makes the step avoiding uploading the same directory twice, when specified both as a Build Artifact and Pipeline File. In such situation the dir will be uploaded only as a Pipeline File.

### Changes

- Introduce `ZipComparator` that can report wether 2 zip files are identical or not
- Use the new `ZipComparator` in `Collector.AddIntermediateFiles` at the end of the function to find duplicated zips 

### Investigation details

The implementation is based on this STO thread: https://stackoverflow.com/a/61113635,
which suggests the following "one-liner" for comparing zip files:

```
diff \
  <(unzip -vqq a.zip  | awk '{$2=""; $3=""; $4=""; $5=""; $6=""; print}' | sort -k3 -f) \
  <(unzip -vqq b.zip  | awk '{$2=""; $3=""; $4=""; $5=""; $6=""; print}' | sort -k3 -f)
```

`unzip -vqq a.zip` lists all the files in the given zip and reports information of them (Length, Method, Size, Date, Time, CRC-32 and Name).

Example output of `unzip -v` (-qq flags just make the output more quiete and only prints the file infos):

```
Archive:  $HOME/test bundles/test_bundle4108757494.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
    7997  Defl:N     1502  81% 01-17-2023 13:30 28c0f1f9  BullsEye_EventuallyFailingInMemoryTests_iphonesimulator15.2-arm64-x86_64.xctestrun

```

`awk '{$2=""; $3=""; $4=""; $5=""; $6=""; print}'` masks out fields that we don't care about (Method, Size, Cmpr, Date and Time).

`sort -k3 -f` sorts the files.

Same solution can be implemented in Go, using the `zip` package.

A zip file can be opened with `zip.OpenReader(pth)`, the returned reader has a `File []*zip.File` property.
`zip.File` has a `zip.FileHeader` field, that stores all the information we need for the comparison:
- Length -> `UncompressedSize64`
- CRC-32 -> `CRC32`
- Name -> `Name`


deployment/collector.go
```diff
@@ -205,3 +212,47 @@ func (c Collector) zipDir(path string) (string, error) {
...
+		}
+
+		if filepath.Ext(item.Path) == ".zip" {
+			zipBuildArtifacts[item.Path] = item
+		} else {
+			mergedDeployableItems = append(mergedDeployableItems, item)
+		}
+	}
+
+	// At this point mergedDeployableItems contains all Pipeline Files and none ZIP Build Artifacts.
```
tothszabi, line 237: Minor nitpick, I would change the `none` to `no` in this sentence. It sounds better with `no` to me. 😄

main.go
```diff
@@ -296,18 +289,40 @@ func collectFilesToDeploy(absDeployPth string, config Config, tmpDir string) (fi
...
-		if err := testResults.Upload(config.AddonAPIToken, config.AddonAPIBaseURL, config.AppSlug, config.BuildSlug); err != nil {
-			log.Warnf("Failed to upload test results: ", err)
-		} else {
-			log.Donef("Success")
+	for i, result := range testResults {
+		if i == 0 {
+			log.Printf("List of test results:")
 		}
+		log.Printf("- %s (generated by the %s)", result.Name, stepNameWithIndex(result.StepInfo))
+	}
```
tothszabi, line 318: This such a great idea. It will help a lot better understanding what is happening in the customer builds. ⭐

test/test.go
```diff
@@ -145,14 +144,24 @@ func ParseTestResults(testsRootDir string) (results Results, err error) {
...
 		// find step-info in dir, continue if no step-info.json as this file is only required if step has exported artifacts also
 		// <root_tests_dir>/<test_dir>/step-info.json
+
+		stepInfoPth := filepath.Join(testDirPath, "step-info.json")
+		if isExists, err := pathutil.IsPathExists(stepInfoPth); err != nil {
+			log.Warnf("Failed to check if step-info.json file exists in dir: %s: %s", testDirPath, err)
+			continue
+		} else if !isExists {
+			continue
+		}
```
tothszabi, line 154: This checks looks much nicer than just trying to immediately read the `step-info.json` file and print out the `Failed to read step-info.json file in dir: %s, error: %s` failure. 

I am just wondering if did you run into this as you were testing your feature? Because it looks a bit out of place when I am looking at the rest of the PR. 😄

test/test.go
```diff
@@ -223,6 +232,16 @@ func ParseTestResults(testsRootDir string) (results Results, err error) {
...
 	for _, result := range results {
+		if len(result.ImagePaths) > 0 {
+			fmt.Printf("Uploading: %s with attachments:", result.Name)
+			for _, pth := range result.ImagePaths {
+				fmt.Printf("- %s", pth)
+			}
+			fmt.Printf("")
+		} else {
+			fmt.Printf("Uploading: %s", result.Name)
+		}
```
tothszabi, line 243: Another nice little addition.

tothszabi (approved):

deployment/collector.go
godrei: Updated in https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/168/commits/ffd454ac2b316395b840ded96e1f061b3ce6c05c

tothszabi (approved):

test/test.go
godrei: Yeah, sorry I haven't mention this in the PR description, I was pretty unhappy with the current log structure and wanted to achieve this log layout:
```
[info] Collecting xyz...
[normal] details

[info] Deploying xyz...
[normal] details
```

That's why I started improving the test result deployment logs, and I faced this `log.Debugf(colorstring.Yellowf`, which seemed to me as a warning log.

I changed it to a warning log, but then I realised there is always 1 directory, withouth test report related files (the test report dir prepared for this step run, the CLI will clean it up after the step run, but when collecting all the test results, it is there) and so that the code always would have print this warning.

Now I check if the required test report file exists and if not, I just continue the loop.

PR #176, opened 2023-03-23
Handle xcarchive parsing errors gracefully

ofalvai:
### Checklist

### Version

### Context

We treat every parsing error as a fatal error that interrupts the upload and makes the entire step fails. This is a bit too drastic, we should at least upload the given file as a plain artifact (without parsed metadata) in this case, so the user can take a look at the problematic file.

### Changes

- Extract xcarchive parsing into a separate function
- Treat the parsing failure as a warning and log the details
- Since metadata is an optimal parameter of the uploading function, we just pass the nil pointer to it if there is no parsed metadata

### Decisions

Fixing all other parser types (APK, AAB, IPA, etc.) is out of scope for this PR. It's just a quick fix to make matters slightly better, but if we see other parsing failures popping up often, let's dedicate some time to fixing all of them.

### DIFF
--- a/uploaders/xcarchiveuploader.go
+++ b/uploaders/xcarchiveuploader.go
@@ -10,6 +10,7 @@ import (
 	"github.com/bitrise-steplib/steps-deploy-to-bitrise-io/deployment"
 )
 
+// DeployXcarchive ...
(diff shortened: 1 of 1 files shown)


ofalvai:
Closing because we need to make server-side changes first, the test runs received an internal server error when uploading the xcarchive without metadata.

PR #178, opened 2023-05-04
Use aab specific package name key

tothszabi:
### Checklist

### Version

### Context

The aab metadata retrieval does not work well with aab files. It tries to fetch the package name the same way as for apk files but the key is different for aabs. For apks the `name` and for aabs the `package` should be used. 

### Changes

- Added a way to differentiate between aok and aab files
- Added a test case to cover the aab path

### DIFF
--- a/androidartifact/apk_info.go
+++ b/androidartifact/apk_info.go
@@ -65,8 +65,13 @@ func parsePackageField(aaptOut, key string) string {
 }
 
 // ParsePackageInfos parses package name, version code and name from `aapt dump badging` command output.
-func ParsePackageInfos(aaptOut string) (string, string, string) {
-	return parsePackageField(aaptOut, "name"),
+func ParsePackageInfos(aaptOut string, isAAB bool) (string, string, string) {
+	packageNameKey := "name"
+	if isAAB {
+		packageNameKey = "package"
+	}
+
+	return parsePackageField(aaptOut, packageNameKey),
 		parsePackageField(aaptOut, "versionCode"),
 		parsePackageField(aaptOut, "versionName")
 }
@@ -144,7 +149,7 @@ func getAPKInfoWithAapt(apkPth string) (ApkInfo, error) {
 	}
 
 	appName := parseAppName(aaptOut)
-	packageName, versionCode, versionName := ParsePackageInfos(aaptOut)
+	packageName, versionCode, versionName := ParsePackageInfos(aaptOut, false)
 	minSDKVersion := parseMinSDKVersion(aaptOut)
 
 	packageContent := ""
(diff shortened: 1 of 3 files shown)


androidartifact/apk_info.go
```diff
@@ -65,8 +65,13 @@ func parsePackageField(aaptOut, key string) string {
 }
 
 // ParsePackageInfos parses package name, version code and name from `aapt dump badging` command output.
-func ParsePackageInfos(aaptOut string) (string, string, string) {
-	return parsePackageField(aaptOut, "name"),
+func ParsePackageInfos(aaptOut string, isAAB bool) (string, string, string) {
```
lpusok, line 68: There is a seperate parseAPKInfo function, is the isAAB paramter needed, or it is always true?

androidartifact/apk_info_test.go
```diff
@@ -14,73 +14,82 @@ import (
...
+			versionName: "1.0",
 		},
 		{
-			name:    "test_without_platformBuildVersionName",
-			aaptOut: `package: name='com.example.birmachera.myapplication' versionCode='1' versionName='1.0'`,
-			want:    "com.example.birmachera.myapplication",
-			want1:   "1",
-			want2:   "1.0",
+			name:        "test_with_platformBuildVersionName",
+			aaptOut:     `package: name='com.example.birmachera.myapplication' versionCode='1' versionName='1.0' platformBuildVersionName='3'`,
```
lpusok, line 40: Which version of the bundletool did you test with? The output of that tool may have changed. We set the version here: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/bbc213a9d8c1ab161c7b31a6fa7660e847e2849c/step.yml#L187.

androidartifact/apk_info.go
tothszabi: We use two functions for the apk parsing which are `parseAPKInfo` and `getAPKInfoWithAapt` one is the fallback for the other:
https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/d00d27bb9961bd22db866568f775a32d975abfe7/androidartifact/apk_info.go#L173-L183

So we need the `isAAB` parameter to differentiate the apk and aab case.

androidartifact/apk_info_test.go
tothszabi: Initially I tried it with 1.11.0 and the latest version `1.14.1`. But after your comment I also checked `1.8.1` and it uses the same `package` key for the package name. 

One thing I would mention is that the preinstalled version:
https://github.com/bitrise-io/bitrise.io/blob/e27882b1a506684f960120b044277a5984fcb82f/system_reports/LINUX/linux-docker-android-20.04.log#L283-L284

And the `1.8.1` default tool version are both pretty old ones. Maybe it is time to upgrade them.

lpusok: Yes, we can just use the preinstalled version to avoid downloading bundletool every time.

PR #180, opened 2023-05-24
Redact secrets in specified files

vshah23:
### Checklist

### Version

### Context

[CI-1502] Feature request in order to be able to redact secret contents of specified files.

### Changes

* Added new field to step.yml
* Added corresponding field to config
* Process the multiline file path input
* Redact files at input paths
* Unit tests

### Investigation details

While investigating this I found the CLI can get secrets from multiple YAML sources:
* Injected as a base64 encoded argument when the CLI is invoked
  * Assuming this is from the secrets tab in the workflow editor
* Via the `.bitrise.secrets.yml`
* Via `.envstore.yml` values which are marked as sensitive

These sources are unmarshalled into `EnvironmentItemOptionsModel` objects. In order to pass the secrets to this step we most likely need to allow the secrets to be accessed by steps using `envman`.

To achieve this I tried to write the combined secrets from the various sources back to the `.bitrise.secrets.yml` file from the CLI and planned to access the secrets in the step using `envman`. The code for writing the secrets back to a file can be found here: https://github.com/vshah23/bitrise/commit/8b98a6632380d91f27fea53163afbaf471ecae85

I found that secrets from the `envstore.yml` source which I marked `is_sensitive: true` in my test file were not included when writing back to the `.bitrise.secrets.yml`.

Also, the vars written to the file by the CLI were marked as `is_sensitive: false` so the step will not know which envs are secrets and which are not unless we make sure to only write the secrets back to the file. The reason for this is the default value for `IsSensitive` in the `EnvironmentItemOptionsModel` is `false` ([bitrise-io/envman/blob/master/models/models_methods.go#L22](https://github.com/bitrise-io/envman/blob/master/models/models_methods.go#L22))

### Not included in this PR

* Handling the secret input from the CLI

### Dependencies

https://github.com/bitrise-io/go-utils/pull/181 (update this PR once this is released)

[CI-1502]: https://bitrise.atlassian.net/browse/CI-1502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vshah23:
@tothszabi @godrei

PR #181, opened 2023-05-26
Flaky test reports

atanas-bitrise:

### Version
Requires a MINOR [version update](https://semver.org/)

### Context

It is fixing an issue where flaky tests weren't reported properly in the Test Reports Add On or in Bitrise Checks. Reporting flaky tests as failures on Bitrise Checks would block merging.

### Changes

This change is mapping each group of flaky tests into one successful test with duration that is the sum of all flaky runs. 

### Investigation details

The alternative solution is to extend the junit format to account for flakiness and update all the systems that will use the new format.

### Decisions

The status of a flaky test group is "success". The duration of the test is reported as the sum of all runs of the flaky test.

### DIFF
--- a/test/converters/xcresult3/converter.go
+++ b/test/converters/xcresult3/converter.go
@@ -151,22 +151,30 @@ func genTestSuite(name string,
 		mtx             sync.RWMutex
 	)
 
+	decoratedTests := decorateTests(tests) 
+
 	testSuite := junit.TestSuite{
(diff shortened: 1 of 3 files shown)


tothszabi:
I am closing this PR because in the meantime we have introduced a solution to this:
https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/190

PR #182, opened 2023-06-30
[RA-1435] Read aab signature

tothszabi:
### Checklist

### Version

### Context

The Release Automation team needs to differentiate between signed and unsigned aab files. At the moment such information is not included in the metadata file while the iOS counterpart has. 

### Changes

The Step extract the cretificate's DN value from aab. The signed aabs always have this while for the unsigned ones it will be empty. It uses the `jarsigner` tool for this which comes as part of the Android SDK.

The change is not covered with unit test because at the moment none of the metadata is tested. We have integration tests to upload the various supported files, unit tests for test result conversions to xml file and extensive tests for the test result handling. It is not clear how the metadata file can be downloaded as it is only used by the website.

### DIFF
--- a/androidsignature/signature.go
+++ b/androidsignature/signature.go
@@ -24,10 +24,11 @@ func Read(path string) (string, error) {
 	if strings.Contains(output, validSignatureMessage) {
 		// The signature details appear in the output in the following format:
 		// - Signed by "C=Aa, ST=Bbbbb, L=Ccccc, O=Ddddd, OU=Eeeee, CN=Fffff"
-		regex := regexp.MustCompile(`- Signed by "(.*?)"`)
-		matches := regex.FindStringSubmatch(output)
-		if len(matches) == 2 {
-			signature = matches[1]
+		regex := regexp.MustCompile(`- Signed by ".*"`)
+		sig := regex.FindString(output)
+		if sig != "" {
+			signature = strings.TrimPrefix(sig, "- Signed by \"")
+			signature = strings.TrimSuffix(signature, "\"")
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


uploaders/aabuploader.go
```diff
@@ -68,6 +69,12 @@ func DeployAAB(item deployment.DeployableItem, artifacts []string, buildURL, tok
 		"build_type":      info.BuildType,
 	}
 
+	signature, err := androidsignature.Read(pth)
+	if err != nil {
+		log.Errorf("Failed to read signature: %s", err)
```
godrei, line 74: This is more like a warning because this won't fail the step.

androidsignature/signature.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	var signature string
+
+	if strings.Contains(output, validSignatureMessage) {
+		// The signature details appear in the output in the following format:
+		// - Signed by "C=Aa, ST=Bbbbb, L=Ccccc, O=Ddddd, OU=Eeeee, CN=Fffff"
+		regex := regexp.MustCompile(`- Signed by "(.*?)"`)
```
godrei, line 27: Wouldn't it be enough to simply remove the suffix (`- Signed by`) from the relevant line and return the rest?

uploaders/aabuploader.go
tothszabi: I copied this from a few lines below so I fixed it at both places.

androidsignature/signature.go
tothszabi: The part we are interested in is also between `""` but I still think your suggestion is better because I think it is easier to process prefix and suffix removal than the understand why there is an array with two items and only the last one is needed.

godrei (approved):

PR #184, opened 2023-09-07
Ci 1502

tothszabi:
### Checklist

### Version

### Context

This PR is mostly shows the great effort Vik did to implement secret redaction in files before upload. 

There is a new input for files which need to be redacted. The steps receives the secret keys list from the CLI and then gets their actual value. Once that is done it will start redacting the files.

### Changes

- There is a new input called `files_to_redact`
- Most of the logic is in the `fileredactor.go` file but the actual redaction related items are coming from the `go-utils` package
- There are unit and e2e tests for the redaction logic


bitrise.yml
```diff
@@ -18,6 +18,15 @@ workflows:
   test:
     steps:
     - go-list: {}
+    - script:
```
ofalvai, line 21: nit: another way to do this: https://github.com/bitrise-steplib/steps-check/commit/474fb5639e6e5b7f7862f38f0ea087bdc65176f4

fileredactor/filepathprocessor.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/bitrise-io/go-utils/v2/env"
+	"github.com/bitrise-io/go-utils/v2/pathutil"
+)
+
+// FilePathProcessor is an interface for an entity which accepts file paths separated by the
```
ofalvai, line 11: It also does env var expansion, I think it should be mentioned in its description.

fileredactor/filepathprocessor.go
```diff
@@ -0,0 +1,75 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		isDir, err := f.pathChecker.IsDirExists(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if path (%s) is a directory: %w", path, err)
+		}
+		if isDir {
+			return nil, fmt.Errorf("path (%s) is a directory and cannot be redacted, please make sure to only provide filepaths as inputs", path)
```
ofalvai, line 68: Is it a deliberate design decision that if one item is a dir, the entire function returns with an error?

step.yml
```diff
@@ -274,6 +274,20 @@ inputs:
         so that the **Deploy to Bitrise.io** Step can build the required custom output for the permanent download URL.  
       is_required: true
       is_expand: false
+  - files_to_redact:
+    opts:
+      title: Files to redact
+      description: |-
+        A newline (`\n`) separated list of file paths to redact secrets from before the step deploys them.
```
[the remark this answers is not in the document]
tothszabi: Good idea. I have added an extra statement to the bottom of the description.

bitrise.yml
tothszabi: This is much better. 👍🏻

fileredactor/filepathprocessor.go
tothszabi: Good catch that env var expansion is actually not needed as the CLI does it. And we are in control of this input so env var expansion will not be disabled. So I have removed that piece of code.

fileredactor/filepathprocessor.go
tothszabi: The recursive directory walking indeed sounds cool but I think it raises a couple of questions.

The redaction only works with text files so if someone uses a directory path then we need to collect all of the text files. We cannot do that based on file extension but I saw some tools which can tell you what kind of type a file is. 
Then there is a risk that users could use a single folder as the input (for example the deploy dir) and just dump everything in there. Even files which does not need redaction. This will increase the runtime of the step unnecessarily. 
Based on the above I think if we force the user to actually list all of the files is better because we do not need to add more logic to the step and also prevent the user making an error and accidentally increasing the step runtime.

And if the users really like this feature and come with feature request where they want to use folders as input value then we can always revisit this and introduce it as a non breaking change.

fileredactor/fileredactor.go
```diff
@@ -0,0 +1,77 @@
...
+		if err := destination.Close(); err != nil {
+			logger.Warnf("Failed to close file: %s", err)
+		}
+	}()
+
+	redactWriter := redactwriter.New(secrets, destination, logger)
+	if _, err := io.Copy(redactWriter, source); err != nil {
+		return err
+	}
+
```
lpusok, line 69: This might be needed:
```suggestion
       if err := redactWriter.Close(); err != nil {
		return fmt.Errorf("failed to redact secrets: %w", err)
       }
```

fileredactor/filepathprocessor.go
ofalvai: I see, I forgot that the input could be non-text file too. I'm okay with the current implementation then!

ofalvai (approved):

fileredactor/fileredactor.go
tothszabi: Great catch. I forgot to close that writer. Added your comment and also updated the error messages a bit.

lpusok (approved):

PR #186, opened 2023-09-22
Add parallel upload

tothszabi:
### Checklist

### Version

### Context

We discovered that uploading a large amount of assets with the steps takes too long. For example, uploading 1200 images takes approximately 15 minutes. This is because the step performs the upload item by item. Adding concurrency to the file uploads will make the step 8x faster. 

### Changes

The main change is that we can control the concurrency with the `BITRISE_DEPLOY_UPLOAD_CONCURRENCY` env var. This is needed because we want to experiment with what is ideal value for our systems. If the env var is missing then the step will use a concurrency of 1. Also at the moment we limit the max value to prevent the anything abusing this value. 

The second change is that the previous implementation was uploading the apks first then everything else. Now the apk upload is merged together with the everything else part.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,7 +6,9 @@ import (
 	"html/template"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/bitrise-io/bitrise/models"
 	"github.com/bitrise-io/envman/envman"
@@ -50,6 +52,7 @@ type Config struct {
 	FilesToRedact                 string `env:"files_to_redact"`
 	DebugMode                     bool   `env:"debug_mode,opt[true,false]"`
 	BundletoolVersion             string `env:"bundletool_version,required"`
+	UploadConcurrency             string `env:"BITRISE_DEPLOY_UPLOAD_CONCURRENCY"`
 }
 
 // PublicInstallPage ...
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -405,77 +414,69 @@ func deploy(deployableItems []deployment.DeployableItem, config Config) (Artifac
...
 
-		artifactURLs, err := uploaders.DeployAPK(apk, androidArtifacts, config.BuildURL, config.APIToken, config.NotifyUserGroups, config.NotifyEmailList, config.IsPublicPageEnabled)
-		if err != nil {
-			errorCollection = handleDeploymentFailureError(err, errorCollection)
-			continue
-		}
+	concurrency := determineConcurrency(config)
+	jobs := make(chan bool, concurrency)
+	combinedItems := append(append(apks, aabs...), others...)
+	mapLock := &sync.RWMutex{}
```
ofalvai, line 423: nit: I'll leave it to you, but maybe `sync.Map` is a better choice here? https://pkg.go.dev/sync#Map

main.go
```diff
@@ -50,6 +52,7 @@ type Config struct {
 	FilesToRedact                 string `env:"files_to_redact"`
 	DebugMode                     bool   `env:"debug_mode,opt[true,false]"`
 	BundletoolVersion             string `env:"bundletool_version,required"`
+	UploadConcurrency             string `env:"BITRISE_DEPLOY_UPLOAD_CONCURRENCY"`
```
ofalvai, line 55: What do you think about generalizing this env var for all steps? Now that I'm facing a very similar problem in the Appcenter upload code, I think it would make sense to have a single env var that controls all step upload concurrencies in builds. This would give a tool to individual users experiencing network flakiness.

ofalvai (approved):

main.go
tothszabi: Thanks for the tip as I did not know about this and I was directly searching for locks so google did not reward me with this result. 😄 

In this case I am leaning towards keeping the lock instead. My reasoning is that if I use `sync.Map` then the rest of the code also needs to be updated and also the rest of the code does not have a concurrent behaviour. 
The problem is with this struct:
```
artifactURLCollection := ArtifactURLCollection{
	PublicInstallPageURLs: map[string]string{},
	PermanentDownloadURLs: map[string]string{},
}
```
which is consumed everywhere in the subsequent stages. So based on these I think this time I will keep the lock.

main.go
tothszabi: We use this as an experiment to determine the right value for the step without killing our network bandwidth. If other step would use this then we could not get accurate results from an experiment because we would control multiple variable part at the same time. Also different steps might have different upload characteristics so they might need different values. 

On the other hand I could see a potential in such a value. And actually as I was writing this message I think I iterated on this feature and have an improved idea. 😄

Based on everything I wrote in the first section I think what would be great is if we could easily add new build level env vars without modifying the backend. Like an option in Optimus Prime where we can add new env vars and also we could specify which experiment is backing that env var. That way we could easily tweak for example the concurrency of every step individually.

main.go
ofalvai: Oh that makes sense, let's keep the lock then for now.

PR #190, opened 2023-12-13
[CI-2284] Add configuration hash to the xml data

tothszabi:
### Checklist

### Version

### Context

The generated xml data representation is not accurate for retried tests. These test runs appear as separate entries in the xml and we need to apply grouping/filtering logic to consume them. 

A simple logic is to simply use the test class name and test name. This works for the regular use cases but there is a new feature which breaks this which is the device configuration tests. This is a handy new Xcode feature where you can tell Xcode to launch the given tests on all language / orientation / interface mode combinations with:
```
override class var runsForEachTargetApplicationUIConfiguration: Bool {
    true
}
```
But tests launched like this will have the same name and identifier and only their configuration will be different. Xcode uses this configuration combo to uniquely identify them. The deploy step was not extracting these and now it will create a checksum from the configuration. 

The backend side grouping can then reliably identify and group related device configuration test runs.

Here is an example of a configuration object extracted from the xcresult file:
```
"configuration" : {
  "_type" : {
    "_name" : "ActionTestConfiguration"
  },
  "values" : {
    "_type" : {
      "_name" : "SortedKeyValueArray"
    },
    "storage" : {
      "_type" : {
        "_name" : "Array"
      },
      "_values" : [
        {
          "_type" : {
            "_name" : "SortedKeyValueArrayPair"
          },
          "key" : {
            "_type" : {
              "_name" : "String"
            },
            "_value" : "XCUIAppearanceMode"
          },
          "value" : {
            "_type" : {
              "_name" : "String"
            },
            "_value" : "1"
          }
        },
        {
          "_type" : {
            "_name" : "SortedKeyValueArrayPair"
          },
          "key" : {
            "_type" : {
              "_name" : "String"
            },
            "_value" : "XCUIDeviceOrientation"
          },
          "value" : {
            "_type" : {
              "_name" : "String"
            },
            "_value" : "1"
          }
        }
      ]
    }
  }
},
```

### Changes

There is a new `ConfigurationHash` field on the xml data struct. This field will be populated by a checksum of the actual configuration taken from the xcresult file.

### DIFF
--- a/test/converters/xcresult3/converter.go
+++ b/test/converters/xcresult3/converter.go
@@ -9,12 +9,13 @@ import (
 	"sync"
 	"time"
 
+	"howett.net/plist"
+
 	"github.com/bitrise-io/go-utils/fileutil"
 	"github.com/bitrise-io/go-utils/log"
 	"github.com/bitrise-io/go-utils/pathutil"
 	"github.com/bitrise-io/go-xcode/xcodeproject/serialized"
 	"github.com/bitrise-steplib/steps-deploy-to-bitrise-io/test/junit"
-	"howett.net/plist"
 )
 
 // Converter ...
@@ -200,15 +201,15 @@ func genTestCase(test ActionTestSummaryGroup, xcresultPath, testResultDir string
 		}
 	}
 
+	testSummary, err := test.loadActionTestSummary(xcresultPath)
+	if err != nil {
+		return junit.TestCase{}, err
+	}
+
 	var failure *junit.Failure
 	var skipped *junit.Skipped
 	switch test.TestStatus.Value {
 	case "Failure":
-		testSummary, err := test.loadActionTestSummary(xcresultPath)
-		if err != nil {
-			return junit.TestCase{}, err
-		}
-
 		failureMessage := ""
 		for _, aTestFailureSummary := range testSummary.FailureSummaries.Values {
 			file := aTestFailureSummary.FileName.Value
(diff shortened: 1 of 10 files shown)


test/converters/xcresult3/converter.go
```diff
@@ -200,15 +201,20 @@ func genTestCase(test ActionTestSummaryGroup, xcresultPath, testResultDir string
 		}
 	}
 
+	testSummary, err := test.loadActionTestSummary(xcresultPath)
```
tothszabi, line 204: Use a `SummaryNotFoundError` and compare it against that.

test/testdata/ios_xml_output.golden
```diff
@@ -1,40 +1,40 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <testsuites>
  <testsuite name="BitriseBasicUITest" tests="4" failures="0" skipped="0" errors="0" time="77.06086790561676">
-  <testcase name="testCollectionView()" classname="BitriseBasicUITest" time="23.776288986206055"></testcase>
-  <testcase name="testInReadTopView()" classname="BitriseBasicUITest" time="34.97628891468048"></testcase>
-  <testcase name="testInterstitialView()" classname="BitriseBasicUITest" time="10.270301938056946"></testcase>
-  <testcase name="testRewardView()" classname="BitriseBasicUITest" time="8.037988066673279"></testcase>
+  <testcase configuration-hash="" name="testCollectionView()" classname="BitriseBasicUITest" time="23.776288986206055"></testcase>
```
tothszabi, line 4: Create a test case with the configuration hash.

lpusok (approved):
Looks ok as far as I can tell.
It seems we reached (already before?) the limits of jUnit format, as far as I can tell configuration-hash is not a standard junit property (https://github.com/testmoapp/junitxml). But it seems enough to tell test cases apart. Do we need the exact orientation etc. displayed? If yes then obviously it will not be enough. For "standard" junit format we could use test properties possibly https://github.com/testmoapp/junitxml#properties-for-suites-and-cases.

Is there an explicit marker to find retried tests (besides the test name)? There could be other edge cases introduced that could break the heuristic to identify retried test.

PR #196, opened 2024-02-19
Handle nonexistent deploy path

ofalvai:
### Checklist

### Version

### Context

When `$BITRISE_DEPLOY_DIR` overwritten with a nonexistent dir and no artifact is added to that dir, the step assumes this path is a file and tries to upload it.

### Changes

Fix the incorrect assumption by checking if the path exists first.


main.go
```diff
@@ -334,9 +334,18 @@ func clearDeployFiles(filesToDeploy []string) []string {
 }
 
 func collectFilesToDeploy(absDeployPth string, config Config, tmpDir string) (filesToDeploy []string, err error) {
+	pathExists, err := pathutil.IsPathExists(absDeployPth)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if %s exists, error: %s", absDeployPth, err)
```
lpusok, line 339:
```suggestion
		return nil, fmt.Errorf("failed to check if %s exists: %s", absDeployPth, err)
```

main.go
```diff
@@ -334,9 +334,18 @@ func clearDeployFiles(filesToDeploy []string) []string {
...
+	}
+	if !pathExists {
+		log.Warnf("Nothing to deploy at %s", absDeployPth)
+		return
+	}
+	
 	isDeployPathDir, err := pathutil.IsDirExists(absDeployPth)
 	if err != nil {
-		return nil, fmt.Errorf("failed to check if DeployPath (%s) is a directory or a file, error: %s", absDeployPth, err)
+		return nil, fmt.Errorf("failed to check if %s is a directory or a file, error: %s", absDeployPth, err)
```
lpusok, line 348:
```suggestion
		return nil, fmt.Errorf("failed to check if %s is a directory or a file: %s", absDeployPth, err)
```

lpusok:
Can we check how often the "is a directory or a file" error happened? By checking a few examples we could be more confident that changing this behavior is not a problem (not a breaking change technically, yes).
As for a future idea: we could expose the warning as a build annotation, as normally the deploy dir not existing is something generally good to know about.

lpusok (dismissed):

PR #197, opened 2024-03-01
Filter invalid reports

tothszabi:
### Checklist

### Version

### Context

At the moment the step and backend does not do any validation if the html report is valid or not. We have only one requirement which is that a report has to have an `index.html` file as its entry point. As a first step this validation is added to the step.

### DIFF
--- a/report/report.go
+++ b/report/report.go
@@ -2,6 +2,7 @@ package report
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/bitrise-io/go-utils/v2/log"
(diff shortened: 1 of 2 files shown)


report/report.go
```diff
@@ -40,14 +41,48 @@ func (h *HTMLReportUploader) DeployReports() []error {
...
 	}
 
-	var errors []error
-	for _, report := range reports {
+	validatedReports, validationErrors := h.validate(reports)
+	if len(validationErrors) != 0 {
+		h.logger.Warnf("Validation errors:\n")
+
+		for _, validationError := range validationErrors {
+			h.logger.Warnf("- %s\n", validationError)
```
zsolt-vicze, line 49: I am not sure about the common practices but a header line might be useful?

report/report.go
```diff
@@ -40,14 +41,48 @@ func (h *HTMLReportUploader) DeployReports() []error {
...
+
+func (h *HTMLReportUploader) validate(reports []Report) ([]Report, []error) {
+	var validatedReports []Report
+	var validationErrors []error
+
+	for _, report := range reports {
+		valid := false
+
+		for _, asset := range report.Assets {
+			if strings.HasSuffix(asset.Path, "index.html") {
```
zsolt-vicze, line 71: Would `/my/path/my_report_index.html` pass this? We need exact "index.html".

tothszabi: Yeah, it would. Fixed it in the next commit.

report/report.go
tothszabi: I did not add a header line because the validation happens after listing all of the found reports here:
https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/2a49db699e82687b19591ebad80121cab539b03d/report/report.go#L32-L51

So my thinking was that we first list all of the reports and then we list the validation errors in the same header block because they belong together. I think helps them link together.

zsolt-vicze (approved):

PR #200, opened 2024-05-30
Provide more artifact information

gaborpongracz:
### Checklist

### Version

### Context/Changes
This change allows us to show more artifact details when we are deploying AABs:
- The app_name is now parsed from the manifest (or from the resources).
- The minSDKVersion is now parsed from the manifest.

From now on the installable artifacts will receive the new artifacts Details page URL and we make it available via the BITRISE_ARTIFACT_DETAILS_PAGE_URL(_MAP) environment variables.


README.md
```diff
@@ -152,6 +153,8 @@ This metadata will be saved with the individual files and restored by the [Pull
...
+| `BITRISE_ARTIFACT_DETAILS_PAGE_URL_MAP` | Details Page URLs by the artifact's path. The default format is `KEY1=>VALUE\|KEY2=>VALUE` but is controlled by the `details_page_url_map_format` input  Examples:  - $BITRISE_DEPLOY_DIR/ios_app.ipa=>https://app.bitrise.io/apps/ios_app/installable-artifacts/ipa-slug - $BITRISE_DEPLOY_DIR/android_app.apk=>https://app.bitrise.io/apps/android_app/installable-artifacts/apk-slug\|$BITRISE_DEPLOY_DIR/ios_app.ipa=>https://app.bitrise.io/apps/ios_app/installable-artifacts/ipa-slug                                                                                                                                                                                                 |
```
BirmacherAkos, line 157: @gaborpongracz What do you think about calling the env `BITRISE_INSTALLABLE_ARTIFACT_DETAILS_PAGE_URL_MAP` it's a bit too long, but at least it highlights that it's only for installable artifacts.

androidartifact/apk_info.go
```diff
@@ -76,6 +81,36 @@ func ParsePackageInfos(aaptOut string, isAAB bool) (string, string, string) {
...
+	return ""
+}
+
+// GetAppNameFromResources parses app name from `aapt dump resources badging` command output.
+func GetAppNameFromResources(aaptOut string) string {
+	pattern := `['"](.*?)['"]`
+	re := regexp.MustCompile(pattern)
+	matches := re.FindAllStringSubmatch(aaptOut, 2)
+	if len(matches) == 2 {
+		appNameMatch := matches[len(matches)-1]
```
tothszabi, line 105: At this point you already make sure that `len(matches) == 2` so you can make this more readable by using simply
```
appNameMatch := matches[1]
```

tothszabi:
One more thing. I think you manually edited the readme. You can use a workflow for that:
```
bitrise run generate_readme
```
The only reason I think you did it manually because when I ran it there were modifications to the readme file.

androidartifact/file_name_test.go
```diff
@@ -44,6 +44,7 @@ func TestParseArtifactPath(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
+		tt := tt
```
tothszabi, line 47: I noticed you are rebinding the loop variable to a local variable to avoid the famous Go for loop variable issue. You only need to do this when you are running your tests in parallel but we are not doing that. In our case the sub tests are not running in separate goroutines. You only need to do what you did here when you have this in your tests:
```
func TestParseArtifactPath(t *testing.T) {
    t.Parallel()
    ...
    for _, tt := range tests {
        tt := tt
        t.Run(tt.name, func(t *testing.T) {
             t.Parallel()
             ...
        }
    }
}
```

tothszabi (approved):

PR #204, opened 2024-07-23
Fix file size upload param

ofalvai:
### Checklist

### Version

### Context

This is one attempt at resolving sporadic upload failures.

For some reason, we convert the file size between an `int64` and `float64` (last code change 7 years ago according to git blame 🙃), which could lead to incorrect file sizes when generating a pre-signed URL for uploading the file. We also query the file size twice (a few seconds apart): first when creating the pre-signed URL, and second when we upload the actual bytes to the URL. This is also a possible race condition.

### Changes

- Get the file size only once and pass it around to all API calls
- Get rid of the `float64` <> `int64` conversion
- Use the `go-units` library (which we already use in `go-steputils`) for pretty-printing the artifact size.


test/converters/xcresult3/converter.go
```diff
@@ -34,7 +34,7 @@ func majorVersion(document serialized.Object) (int, error) {
 	if err != nil {
 		return -1, err
 	}
-	return int(major.(uint64)), nil
+	return int(major.(uint64)), err
```
tothszabi, line 37: nit: I would leave it as `nil` to better articulate that at this point there should be no error. `err` would be nil as well here so technically is correct. But for me the `nil` one reads better.

ofalvai: Sorry, this is a leftover change from previous commits. This is the line that the old `errcheck` linter was complaining about (before merging #205), and I tried this change to see if this is the issue. It was still complaining, and now I reverted the change

tothszabi (approved):

PR #205, opened 2024-07-23
Lint codebase, fix CI

ofalvai:
### Checklist

### Version

### Context

A sudden lint failure in #204, and I couldn't figure out what it was trying to tell me, so I decided it's time to apply our org-wide linters to this repo too.

### Changes

- Fix golangci-lint violations
- Fix YAML formatting violations
- Fix some step.yml schema errors reported by the validator

### Investigation details

Go lint violations for extra context:

```
+ golangci-lint run --timeout 5m --color always
androidartifact/file_name_test.go:272:6: func `comparseSlice` is unused (unused)
func comparseSlice(a, b []string) bool {
     ^
androidartifact/file_name_test.go:285:6: func `compareMapStringStringSlice` is unused (unused)
func compareMapStringStringSlice(a, b map[string][]string) bool {
     ^
androidartifact/file_name_test.go:429:7: type `args` is unused (unused)
        type args struct {
             ^
androidartifact/apk_info.go:88:22: S1030: should use manifestContent.String() instead of string(manifestContent.Bytes()) (gosimple)
                RawPackageContent: string(manifestContent.Bytes()),
                                   ^
androidartifact/file_name.go:205:4: SA9003: empty branch (staticcheck)
                        if len(artifact.APK) != 0 {
                        ^
report/collect.go:121:5: S1002: should omit comparison to bool constant, can be simplified to `!strings.HasPrefix(contentType, plainContentType)` (gosimple)
        if strings.HasPrefix(contentType, plainContentType) == false {
           ^
report/collect.go:98:3: SA9003: empty branch (staticcheck)
                if err := file.Close(); err != nil {
                ^
report/api/api_test.go:93:41: SA1008: keys in http.Header are canonicalized, "BUILD_API_TOKEN" is not canonical; fix the constant or use http.CanonicalHeaderKey (staticcheck)
                        assert.Equal(t, []string{authToken}, request.Header["BUILD_API_TOKEN"])
                                                             ^
report/api/api_test.go:149:41: SA1008: keys in http.Header are canonicalized, "BUILD_API_TOKEN" is not canonical; fix the constant or use http.CanonicalHeaderKey (staticcheck)
                        assert.Equal(t, []string{authToken}, request.Header["BUILD_API_TOKEN"])
                                                             ^
fileredactor/fileredactor.go:13:7: const `bufferSize` is unused (unused)
const bufferSize = 64 * 1024
      ^
test/converters/xcresult/xcresult.go:26:82: S1002: should omit comparison to bool constant, can be simplified to `!exist` (gosimple)
                        if exist, err := pathutil.IsPathExists(testSummariesPlistPath); err != nil || exist == false {
                                                                                                      ^
test/test_test.go:66:5: testinggoroutine: call to (*T).Fatal from a non-test goroutine (govet)
                                t.Fatal("app_slug must be specified")
                                ^
test/test_test.go:69:5: testinggoroutine: call to (*T).Fatal from a non-test goroutine (govet)
                                t.Fatal("build_slug must be specified")
                                ^
test/test_test.go:74:5: testinggoroutine: call to (*T).Fatal from a non-test goroutine (govet)
                                t.Fatal("failed to execute get request, error:", err)
                                ^
test/test_test.go:314:8: ineffectual assignment to err (ineffassign)
        want, err := fileutil.ReadStringFromFile(filepath.Join("testdata", "ios_device_config_xml_output.golden"))
              ^
test/test_test.go:60:2: SA2002: the goroutine calls T.Fatal, which must be called in the same goroutine as the test (staticcheck)
        go func() {
```


androidartifact/file_name_test.go
```diff
@@ -269,38 +268,6 @@ func Test_mapBuildArtifacts(t *testing.T) {
 	}
 }
 
-func comparseSlice(a, b []string) bool {
```
ofalvai, line 272: Unused code

report/collect.go
```diff
@@ -96,7 +96,7 @@ func detectContentType(path string) string {
 	}
 	defer func() {
 		if err := file.Close(); err != nil {
-			// This is empty on purpose to please the linter
+			panic(err)
```
tothszabi, line 99: This alter the behaviour significantly because it will fail the step. Previously the worst case was that we most probably just leaked some memory. 

What if we just simply do 
```
_ = file.Close()
```
and annotate it so the linter does not complain about it?

ofalvai: Good point, thank you! The leak probably doesn't matter in these short-lived step processes anyway.

tothszabi (approved):

PR #206, opened 2024-08-01
[RA-3066] Use libs for artifact meta data parsing

gaborpongracz:
### Checklist

### Version

### Context
We want to use the same metadata parsing logic in the Deploy step and on the site.

### Changes
- Move the parsing logic to the proper libs (go-android, go-xcode, go-utils)


main.go
```diff
@@ -10,7 +10,10 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/bitrise-steplib/steps-deploy-to-bitrise-io/bundletool"
+	androidparser "github.com/bitrise-io/go-android/v2/metaparser"
+	iosparser "github.com/bitrise-io/go-xcode/v2/metaparser"
+
```
godrei, line 15: nit: unnecessary empty line

uploaders/aabuploader.go
```diff
@@ -3,89 +3,54 @@ package uploaders
...
-	signature, err := androidsignature.Read(pth)
-	if err != nil {
-		log.Warnf("Failed to read signature: %s", err)
+	if aabInfo.AppInfo.AppName == "" {
+		u.logger.Warnf("App name is undefined, AndroidManifest.xml package content:\n%s", aabInfo.AppInfo.RawPackageContent)
 	}
-	aabInfoMap["signed_by"] = signature
 
-	splitMeta, err := androidartifact.CreateSplitArtifactMeta(pth, artifacts)
+	splitMeta, err := androidartifact.CreateSplitArtifactMeta(u.logger, pth, artifacts)
```
godrei, line 41: Why isn't the split meta creation part of the lib?

uploaders/xcarchiveuploader.go
```diff
@@ -3,38 +3,30 @@ package uploaders
 import (
 	"fmt"
 
-	logV2 "github.com/bitrise-io/go-utils/v2/log"
-	"github.com/bitrise-io/go-xcode/v2/artifacts"
-	"github.com/bitrise-io/go-xcode/v2/zip"
+	"github.com/pkg/errors"
```
godrei, line 6: pls use the stdlib's errors package

uploaders/aabuploader.go
gaborpongracz: The main artifact parser logic which is purely moved to the libs is about only individual artifacts. Split metadata uses multiple artifacts which is also a client-side decision. For now, I prefer to leave this to the deploy step, because on the Release Management side, we will not use this feature. We will probably support AAB over-the-air installation in our way.

godrei (approved):

PR #207, opened 2024-08-02
Deploy the snapshot of generic files under 1GB file size

godrei:
### Checklist

### Version

### Context

This PR implements a workaround for deploying generic files, that are potentially being modified during the deployment process.
The workaround is creating a snapshot of generic files at max 1GB file size, by copying them to a tmp path and deploying this tmp file, instead of the original file.

### DIFF
--- a/uploaders/fileuploader.go
+++ b/uploaders/fileuploader.go
@@ -2,20 +2,45 @@ package uploaders
 
 import (
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/bitrise-io/go-utils/log"
+	"github.com/bitrise-io/go-utils/pathutil"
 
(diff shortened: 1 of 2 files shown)


ofalvai (approved):

uploaders/fileuploader.go
```diff
@@ -2,20 +2,54 @@ package uploaders
...
+					log.Warnf("Failed to remove snapshot file: %s", err)
+				}
+			}()
+			pth = snapshotPth
+			deploySnapshot = true
+		}
+	}
+
+	if deploySnapshot {
+		log.Printf("Deploying file snapshot: %s", pth)
```
ofalvai, line 43:
```suggestion
		log.Printf("Deploying snapshot of original file: %s", pth)
```
How about this? I think it provides a bit more context to users who don't know the behind-the-scenes stuff

ofalvai (approved):

PR #212, opened 2025-01-31
Don't try to deploy nonexistent files

ofalvai:
### Checklist

### Version

### DIFF
--- a/uploaders/apkuploader.go
+++ b/uploaders/apkuploader.go
@@ -16,7 +16,7 @@ func (u *Uploader) DeployAPK(item deployment.DeployableItem, artifacts []string,
 		return ArtifactURLs{}, err
 	}
 
-	u.logger.Printf("apk infos: %+v", printableAppInfo(apkInfo.AppInfo))
+	u.logger.Printf("APK metadata: %+v", printableAppInfo(apkInfo.AppInfo))
 
 	splitMeta, err := androidartifact.CreateSplitArtifactMeta(u.logger, pth, artifacts)
 	if err != nil {


Badlazzor (approved):

main.go
```diff
@@ -340,6 +341,11 @@ func logDeployFiles(files []deployment.DeployableItem, logger loggerV2.Logger) {
 func clearDeployFiles(filesToDeploy []string, logger loggerV2.Logger) []string {
```
tothszabi, line 341: Should not we perform the file existence check earlier?

At the moment the step output looks like this:
```
List of files to deploy
- /bitrise/src/Apps/app-stg-release-unsigned.apk

Deploying files
Uploading apk file: /bitrise/src/Apps/app-stg-release-unsigned.apk
analyzing apk
Failed to parse APK info: failed to unzip the APK, error: open /bitrise/src/Apps/app-stg-release-unsigned.apk: no such file or directory
deploy failed, error: failed to get apk infos, output: asset W 01-31 12:06:07  2347  2347] Asset path /bitrise/src/Apps/app-stg-release-unsigned.apk is neither a directory nor file (type=1).
ERROR: dump failed because assets could not be loaded, error: exit status 1
deploy failed, error: failed to get apk infos, output: asset W 01-31 12:06:07  2347  2347] Asset path /bitrise/src/Apps/app-stg-release-unsigned.apk is neither a directory nor file (type=1).
ERROR: dump failed because assets could not be loaded, error: exit status 1
```

The step lists that it will deploy the `/bitrise/src/Apps/app-stg-release-unsigned.apk` file and then later it states it could not deploy the file. Why did it find the file in the first place then?

I would say that we should check if the file exists when we collect the deployable files.

ofalvai: That was my thinking as well! `clearDeployFiles()` is called right after collecting the files based on the step inputs: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/a94282c38d35c1e034dce1c7984009d4247e9fd3/main.go#L146-L150

And it's called before printing the list of deployable files (L168)

tothszabi: Yeah, I was thinking of doing it earlier like in the `collectFilesToDeploy` method. My thinking was that why are we even finding files which do not exist. The clear function was created more for filtering out the useless files.

PR #214, opened 2025-03-19
[CI-4420] New test result extraction

tothszabi:
### Checklist

### Version

### Context

Since Xcode 16 we have a new way to extract test information from the xcresult file with xcresulttool. This PR removes the legacy flag and updates the step to use the new method. 

### Investigation details

The new method uses a completely different data structure and cannot be easily fit into the old one. So there are a new set data models and the whole method is separated from the old one.

### DIFF
--- a/report/collect_test.go
+++ b/report/collect_test.go
@@ -23,7 +23,7 @@ func TestSpecialContentTypes(t *testing.T) {
 		{
 			name:     "Javascript file",
 			fileName: "abc.js",
-			expected: "application/javascript",
+			expected: "text/javascript; charset=utf-8",
 		},
 		{
 			name:     "CSS file",
--- a/test/converters/converters_test.go
+++ b/test/converters/converters_test.go
@@ -27,17 +27,17 @@ func TestXCresult3Converters(t *testing.T) {
 				Tests:    2,
 				Failures: 0,
 				Errors:   0,
-				Time:     0.26388001441955566,
+				Time:     0.26063,
 				TestCases: []junit.TestCase{
 					{ // plain test case
 						Name:      "testExample()",
 						ClassName: "rtgtrghtrgTests",
-						Time:      0.0006339550018310547,
+						Time:      0.00063,
 					},
 					{ // plain test case
 						Name:      "testPerformanceExample()",
 						ClassName: "rtgtrghtrgTests",
-						Time:      0.2632460594177246,
+						Time:      0.26,
 					},
 				},
 			},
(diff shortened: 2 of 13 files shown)


report/collect_test.go
```diff
@@ -23,7 +23,7 @@ func TestSpecialContentTypes(t *testing.T) {
 		{
 			name:     "Javascript file",
 			fileName: "abc.js",
-			expected: "application/javascript",
+			expected: "text/javascript; charset=utf-8",
```
tothszabi, line 26: The newer Go stdlib outputs this as the content type for js files.

tothszabi, line 26: We used an old stack and the new Go runtime uses this content type for JS files.

test/converters/xcresult3/model3/conversion.go
```diff
@@ -0,0 +1,98 @@
...
+
+		summary.TestPlans = append(summary.TestPlans, testPlan)
+	}
+
+	return &summary, nil
+}
+
+func extractDuration(text string) time.Duration {
+	// Duration is in the format "123.456789s"
+	duration, err := time.ParseDuration(text)
```
tothszabi, line 69: The test information we get back uses human readable strings for duration instead of some kind of numeric value. That is why you can see a bunch of time related changes in the tests.

test/converters/xcresult3/model3/testresults.go
tothszabi, line 1: We do not need many fields from these data structures. This is the full set of what the xcresulttool returns. I thought I will implement all of them so we can track changes over time and we can also see what is available.

test/converters/xcresult3/model3/testresults.go
```diff
@@ -0,0 +1,63 @@
...
+type TestNode struct {
+	Identifier string       `json:"nodeIdentifier"`
+	Type       TestNodeType `json:"nodeType"`
+	Name       string       `json:"name"`
+	Details    string       `json:"details"`
+	Duration   string       `json:"duration"`
+	Result     TestResult   `json:"result"`
+	Tags       []string     `json:"tags"`
+	Children   []TestNode   `json:"children"`
+}
```
tothszabi, line 58: The Apple devs went with a flexible structure. That means everything is a `TestNode` and it can have children `TestNode`s. And one of this test node can be of 15 different types. So to understand what is the relationship between these types I had to create xcresult files with different scenarios. To give you an example:
If you have a failing test then the child of such a `Test Case` node will be a `Failure Message` node containing the message. But if you have test retry turned on then the children will be of `Repetition` nodes and children of these nodes will contain the `Failure Message` type.

test/converters/xcresult3/model3/conversion.go
```diff
@@ -0,0 +1,98 @@
...
+	lastNode := testNode.Children[childrenCount-1]
+	if lastNode.Type == TestNodeTypeRepetition {
+		return extractFailureMessage(lastNode)
+	}
+
+	failureMessage := ""
+
+	for _, child := range testNode.Children {
+		if child.Type == TestNodeTypeFailureMessage {
+			// The failure message appears in the Name field and not in the Details field.
```
tothszabi, line 92: Other nodes have the useful information in the `Details` but not the `Failure Message` type which uses the `Name` field for this. Hopefully this is not bug in the tool.

test/converters/xcresult3/model3/export.go
```diff
@@ -0,0 +1,35 @@
...
+func (t *Timestamp) UnmarshalJSON(b []byte) error {
+	var timestamp float64
+	if err := json.Unmarshal(b, &timestamp); err != nil {
+		return err
+	}
+
+	*t = Timestamp(time.Unix(int64(timestamp), 0))
+
+	return nil
+}
```
tothszabi, line 35: I wrote above that for the test cases we receive human readable string for duration. Attachments have a proper unix timestamp. We do not use the timestamp at the moment but thought that I will implement everything here as well.

Go cannot parse a unix timstamp by default and requires manual unmarshalling.

test/converters/xcresult3/converter_test.go
```diff
@@ -47,66 +47,60 @@ func TestConverter_XML(t *testing.T) {
...
 					{
 						Name: "testGameStyleSwitch()", ClassName: "BullsEyeUITests",
-						Time: 9.606701016426086,
+						Time: 9,
 					},
 				},
 			},
 			{ // testFlakyFeature() was flaky: failed once, and then passed the second run
-				Name: "BullsEyeFlakyTests", Tests: 3, Failures: 1, Skipped: 1, Errors: 0, Time: 0.22202682495117188,
+				Name: "BullsEyeFlakyTests", Tests: 2, Failures: 0, Skipped: 1, Errors: 0, Time: 0.12,
```
godrei, line 97: `Tests` have changed from 3 to 2, is this intentional? Also, the comment doesn't seem in sync with the test case expectation.

test/converters/xcresult3/model3/conversion.go
```diff
@@ -0,0 +1,98 @@
+package model3
```
godrei, line 1: I kept reading it as mode`13` (thirteen), and it took some time for me to realize that it is model`3` (three) 😂

test/converters/xcresult3/model3/conversion.go
```diff
@@ -0,0 +1,98 @@
...
+			for _, testSuiteNode := range testBundleNode.Children {
+				if testSuiteNode.Type != TestNodeTypeTestSuite {
+					return nil, fmt.Errorf("test suite expected but got: %s", testSuiteNode.Type)
+				}
+
+				testSuite := TestSuite{Name: testSuiteNode.Name}
+
+				for _, testCaseNode := range testSuiteNode.Children {
+					if testCaseNode.Type != TestNodeTypeTestCase {
+						return nil, fmt.Errorf("test case expected but got: %s", testCaseNode.Type)
```
godrei, line 35: These errors made me wonder if we should fall back to the legacy parser in case of an error in the new parser, what do you think of this idea?

test/converters/xcresult3/model3/conversion.go
godrei: Hmm, maybe we should emit analytic logs about unexpected parse results? Like if the child.Name is empty for a FailureMessage node, or if the Details is not empty for this type?

test/converters/xcresult3/model3/testresults.go
```diff
@@ -0,0 +1,63 @@
+package model3
+
+type TestNodeType string
+
+// These are all the types the xcresulttool (version 23500, format version 3.53) supports.
```
godrei, line 5: Maybe as an improvement, we should put together a test, that checks if a new format was shipped with a new Xcode version.

test/converters/xcresult3/xcresulttool.go
```diff
@@ -75,13 +81,24 @@ func xcresulttoolGet(xcresultPth, id string, v interface{}) error {
...
-	args := []string{"xcresulttool", "export", "--path", xcresultPth, "--id", id, "--output-path", outputPth, "--type", "file"}
-	isLegacyFlag, err := isLegacyFlagNeededForXcresulttoolVersion()
+	args := []string{"xcresulttool", "export"}
+
+	supportsNewMethod, err := supportsNewExtractionMethods()
 	if err != nil {
 		return err
 	}
-	if isLegacyFlag {
-		args = append(args, "--legacy")
```
godrei, line 84: Don't we use the `--legacy` flag anymore, not even with the legacy exporter?

test/converters/converters_test.go
```diff
@@ -46,98 +46,98 @@ func TestXCresult3Converters(t *testing.T) {
...
-						Time:      0.031695008277893066,
+						Time:      0.032,
 					},
 					{
 						Name:      "testFailMe",
 						ClassName: "_TtCC17rtgtrghtrgUITests17rtgtrghtrgUITests18rtgtrghtrg3UITests",
-						Time:      0.09049093723297119,
+						Time:      0.09,
 						Failure: &junit.Failure{
-							Value: "/Users/tamaspapik/Develop/ios/rtgtrghtrg/rtgtrghtrgUITests/rtgtrghtrgUITests.swift:68 - XCTAssertTrue failed",
```
godrei, line 62: the new parsing method removes the failing line info. can we get this back from somewhere?

test/converters/xcresult3/converter.go
```diff
@@ -130,6 +150,107 @@ func (c *Converter) XML() (junit.XML, error) {
...
+				log.Warnf("Failed to rename %s to %s", oldPath, newPath)
+			}
+		}
+	}
+
+	if err := os.Remove(manifestPath); err != nil {
+		return err
+	}
+
+	return nil
```
godrei, line 251: Can we move this logic into a new function?

test/converters/xcresult3/converter_test.go
tothszabi: Yes, because the new extraction method handles the retried tests better and only shows the final state for them.

test/converters/xcresult3/model3/conversion.go
tothszabi: Great idea and I made this change.

test/converters/xcresult3/xcresulttool.go
tothszabi: Reintroduced the `legacy` flag as we need it fallback extraction.

test/converters/converters_test.go
tothszabi: I checked the this with more Xcode 16 xcresult files and we do have the file and line number in the error result:
```
DarkAndLightModeTests.swift:25: failed - This test failed
```
And actually we a test case which showcases this:
https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/620564537af982a2638209977be29473d5d29a57/test/converters/xcresult3/converter_test.go#L142

Apart from these I did not find any other way with the new way to get more error details.

What I suspect is that the xcresult file this test uses has an older data structure which maybe the new extraction method cannot fully understand. And could explain why it always works with Xcode 16 generated ones.

test/converters/xcresult3/converter.go
```diff
@@ -99,26 +99,38 @@ func (c *Converter) XML() (junit.XML, error) {
...
-		return parse(c.xcresultPth)
+		junitXml, err := parse(c.xcresultPth)
+		if err == nil {
+			return junitXml, nil
+		}
+
+		log.Warnf(fmt.Sprintf("Failed to parse extraction method: %s", err))
+		log.Warnf("Falling back to legacy extraction method")
+
+		logWarn("xcresult3-parsing", nil, "error: %s", err)
```
godrei, line 115: As far as I can see, this is the only place where we use `logWarn`, and the `data` parameter is nil here. Do we need this function input?

godrei: Also, for the first read `logWarn` doesn't really tell that it sends analytic logs. What about renaming it to `sendAnalytics` or `sendRemoteLog` or `sendRemoteWarning`?

godrei (approved):

PR #217, opened 2025-05-14
[CI-4808] Legacy xcresult parsing input

tothszabi:
### Checklist

### Version

### Context

Some customers xcresult file might not be compatible with the new parsing (or there is a bug in the extraction tool) and this new input can force the step to use the legacy extraction method.

### DIFF
--- a/step.yml
+++ b/step.yml
@@ -336,6 +336,21 @@ inputs:
     value_options:
     - "false"
     - "true"
+- use_legacy_xcresult_extraction_method: "false"
+  opts:
+    category: Debugging
+    title: Legacy extraction method for Xcode test results
+    summary: |-
+      The Step will try to extract the test results with the older extraction method.
+    description: |-
+      The Step will try to extract the test results with the older extraction method.
+    
+      Some Xcode test results are not compatible with the new extraction method and cannot extract the test results. 
+      If you encounter this issue, please set this input to `true` and try again.
+    is_required: true
+    value_options:
+    - "false"
+    - "true"
 outputs:
 - BITRISE_PUBLIC_INSTALL_PAGE_URL:
   opts:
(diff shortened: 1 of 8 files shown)


step.yml
```diff
@@ -336,6 +336,21 @@ inputs:
     value_options:
     - "false"
     - "true"
+- use_legacy_xcresult_extraction_method: "false"
+  opts:
+    category: Debugging
+    title: Legacy extraction method for Xcode test results
+    summary: |-
+      The Step will try to extract the test results with the older extraction method.
```
godrei, line 344: Would it make sense to be more verbose about what legacy/older extraction method means? Like mentioning the related Xcode version. I can imagine new xcresult formats and new extraction methods in the feature.

test/converters/converters.go
```diff
@@ -12,10 +12,11 @@ import (
 	"github.com/bitrise-steplib/steps-deploy-to-bitrise-io/test/junit"
 )
 
-// Intf is the required interface a converter need to match
+// Intf is the required interface a converter needs to match
 type Intf interface {
 	XML() (junit.XML, error)
 	Detect([]string) bool
+	Setup(useOldXCResultExtractionMethod bool)
```
godrei, line 19: Do we need this method for the interface? it is only related to the xcresult converter? Can't we use this information when the xcresult converter is initialised?

tothszabi: We are creating the converters as global package level constants here:
https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/7e83efd64d8060a8141cc40484ca99d17ac3684c/test/converters/converters.go#L22-L26

That means we cannot pass in the user level settings as the processing happens later. So we need to either add a function to the interface and call it before the parsing happens. Or we could try to cast the converter to the `xcresult3` type and when it succeeds we can pass the parameter to it. I choose the first one because it looked nicer.

The ideal would be something as what you wrote that we could customize the individual converter at initialization. I looked at it but it seemed too much of a change for the time being.

step.yml
tothszabi: Great idea. I have updated it.

godrei (approved):

PR #259, opened 2026-03-10
Move testreport to steputils & converters to go-<platform>

gergely-sallai:
### Checklist

### Version

We can get away with a patch version change from an end user perspective. As no behavioral or input/output changes were made. The only "API" change was in a package moving and rename.

### Context
Refactored and moved part of the test capability to supporting libraries in order to eliminate steps depending on this step.

### Changes
- Moved testresult and testasset related code to `go-steputils`
- Moved junit converters to `go-android`
- Moved xcresult converters to `go-xcode`
- Flattened packages in `test` package to simplify after conversion and other tool have been removed.
- Updated dependencies and vendoring


test/converters.go
```diff
@@ -19,13 +19,13 @@ type Converter interface {
...
+var converterList = []Converter{
 	&junitxml.Converter{},
 	&xcresult.Converter{},
 	&xcresult3.Converter{},
 }
 
-// List lists all supported converters
-func List() []Converter {
-	return converters
+// AvailableConverters lists all supported converters
```
ofalvai: nit: we don't enforce function comments in step repos, only in lib repos. Personal preference, but I find these trivial doc comments noise, and I like to remove them when I encounter one.

test/converters.go
ofalvai: Have you thought about moving this interface to `go-steputils` as well? It's a bit strange that the interface is defined in this step, but the implementations are in `go-go-xcode` and `go-android`. But if there are good reasons for this, then I can live with this approach :)

ofalvai: The return value of `Convert()` (`testreport.TestReport`) is already in `go-steputils`

test/converters.go
gergely-sallai: I can remove, it just stayed there.. will update.

test/converters.go
gergely-sallai: No there's no good reason to keep it here. This is perfectly valid comment. I just missed it, and Go doesn't enforce this..

test/converters.go
gergely-sallai: done

test/converters.go
gergely-sallai: done

ofalvai (approved):

PR #261, opened 2026-05-28
[CI-5635] Multipart upload

tothszabi:
### Checklist

### Version

### Context

Introduce multi part upload support to make large file transfers faster. 

The feature is gated behind an env var which means the step still uses single concurrency upload by default. This will help us roll it out gradually.


uploaders/common.go
```diff
@@ -50,284 +37,6 @@ type TransferDetails struct {
 	Hostname string
 }
 
-type UploadTask struct {
```
tothszabi, line 53: These one part upload methods were moved to the `upload_single` file.

uploaders/upload_multipart.go
```diff
@@ -0,0 +1,465 @@
...
+}
+
+// UploadedPart records the ETag returned by S3 after uploading a single part.
+// Both fields are required by complete_multipart_upload on the server side.
+type UploadedPart struct {
+	PartNumber int    `json:"part_number"`
+	ETag       string `json:"etag"`
+}
+
+func createMultipartArtifact(buildURL, token string, artifact ArtifactArgs, artifactType, contentType string, archiveAsArtifact bool, pipelineMeta *deployment.IntermediateFileMetaData, logger logV2.Logger) ([]MultipartUploadTask, error) {
```
godrei, line 94: `createMultipartArtifact` and `createArtifact` seems to be very similar, have you considered sharing code between them?

uploaders/upload_multipart.go
```diff
@@ -0,0 +1,465 @@
...
+		return nil, fmt.Errorf("part_urls count (%d) does not match expected (%d) for file size %d and part size %d", partCount, expectedPartCount, fileSize, partSize)
+	}
+
+	type partResult struct {
+		partNumber int
+		etag       string
+		err        error
+	}
+
+	results := make(chan partResult, partCount)
```
godrei: does partCount always equal to partURLs? If so do we need partCount as a separate parameter?

uploaders/upload_single.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	return ArtifactURLs{
+		PermanentDownloadURL: artifactResponse.PermanentDownloadURL,
+		DetailsPageURL:       artifactResponse.DetailsPageURL,
+		PublicInstallPageURL: artifactResponse.PublicInstallPageURL,
+	}, nil
+}
+
+func (u *Uploader) uploadSingle(buildURL, token string, artifact ArtifactArgs, artifactType, contentType string, item *deployment.DeployableItem, buildArtifactMeta *AppDeploymentMetaData) ([]ArtifactURLs, error) {
```
godrei: NIT: the most important func is at the bottom, I suggest moving it to the top

godrei (commented):
Do you plan to add an e2e test for multipart upload?

uploaders/upload_multipart.go
tothszabi: It is on purpose because I want to clean up the single part upload an only have the multipart one. One of the reason is that the storage provider's multipart endpoints work with a single part as well so we could use the new multipart upload for both cases. 

Originally I completely removed the single part upload from the step but then I added it back as this change has a slight risk factor so I want to introduce it gradually and have a fallback option. The whole feature will be driven by a feature flag.

uploaders/upload_multipart.go
tothszabi: That is correct. I will remove it.

godrei (approved):

PR #267, opened 2026-07-03
Log status code, URL, and response body on failed artifact create requests

ofalvai:
### Checklist

### Version

### Context

We had an incident where a failed artifact upload logged an empty error message, giving us no way to diagnose the failure:

```
failed file deploy:
  failed to create artifact (/tmp/snapshot3554133073/src.zip):

deploy failed, error:
  failed file deploy:
    failed to create artifact (/tmp/snapshot3554133073/src.zip):
```

### Changes

- In `createArtifact` (`upload_single.go`) and `createMultipartArtifact` (`upload_multipart.go`), when the create-artifact request returns a non-200 response whose body doesn't contain a (non-empty) `error_msg` field, fall back to an error that includes the HTTP status code, request URL, and raw response body instead of an empty string.
- Include the request URL in the "failed to perform create artifact request" error as well, so connection-level failures (e.g. a malformed/unreachable URL) are identifiable.

### Investigation details

Root cause: `errors.New(createResponse.ErrorMessage)` returned an error with an empty `Error()` string whenever the server's JSON error response didn't populate `error_msg` (or parsed successfully but with an empty value). This error propagated unchanged through all the `%w`-wrapped errors up to the top-level deploy failure message, producing the truncated log seen in the incident.

### Decisions

This is a minimal, targeted fix for the specific swallowed-error case. A broader follow-up is planned to migrate the upload retry logic in `upload_single.go`/`upload_multipart.go` onto the shared retry HTTP client already used elsewhere in this repo (`report/api/api.go`, `test/test.go`), to unify retry/error-logging behavior across all upload paths.

### DIFF
--- a/uploaders/upload_multipart.go
+++ b/uploaders/upload_multipart.go
@@ -152,7 +152,7 @@ func createMultipartArtifact(buildURL, token string, artifact ArtifactArgs, arti
 
 		response, err = http.DefaultClient.Do(req)
 		if err != nil {
-			return fmt.Errorf("failed to perform create artifact request, error: %s", err)
+			return fmt.Errorf("failed to perform create artifact request (%s), error: %s", uri, err)
 		}
 
 		defer func() {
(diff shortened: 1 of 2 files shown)


uploaders/upload_multipart.go
```diff
@@ -152,7 +152,7 @@ func createMultipartArtifact(buildURL, token string, artifact ArtifactArgs, arti
 
 		response, err = http.DefaultClient.Do(req)
 		if err != nil {
-			return fmt.Errorf("failed to perform create artifact request, error: %s", err)
+			return fmt.Errorf("failed to perform create artifact request (%s), error: %s", uri, err)
```
godrei, line 155: NIT: these duplicated error handling could be shared, it is fine since a bigger change is planned in this area.

godrei (approved):

uploaders/upload_multipart.go
ofalvai: If you mean `upload_single.go` and `upload_multipart.go`: these are new additions by https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/261, so I didn't want to mess with it too much (FYI @tothszabi). I assume the plan is to throw away `upload_single.go` after the full rollout?

PR #270, opened 2026-08-28
Migrate go-utils and go-steputils to v2 (ziputil + helper packages)

godrei:
## What

First step of migrating this Step off the v1 `go-utils` and `go-steputils` libraries. Two logically separate commits:

1. **Migrate ziputil to v2** — switch the Step to `go-utils/v2/ziputil` (`ZipManager`) and bump `go-utils/v2` to the release that contains it.
2. **Migrate helper packages to v2** — the trivial, behaviour-preserving package swaps:
   - `stepconf.Parse` -> `stepconf.NewInputParser(envRepo).Parse`
   - `pathutil` free functions -> `PathProvider` / `PathChecker` / `PathModifier`
   - `fileutil` read/write helpers -> stdlib `os.ReadFile` / `os.WriteFile` (v2 `FileManager` has no read-string/bytes method)
   - `urlutil` and `retry.Times/Wait/Try` -> v2 (identical APIs)
   - `retry.NewHTTPClient` -> `retryhttp.NewClient`
   - `tools.ExportEnvironmentWithEnvman` -> `export.Exporter.ExportOutput`

The v1 `log` package is intentionally left in place; migrating it (and the analytics-backed APK/AAB parse warnings) is a separate follow-up so this PR stays a focused, no-behaviour-change swap.

## Notes

- `go-steputils` v1 is no longer used and is dropped from `go.mod`. `go-utils` v1 stays only for `log`.
- Adding the `go-steputils/v2/export` dependency transitively pulls v1 `go-utils/ziputil`, `filedownloader`, `parseutil` and `retry` into vendor, because the export package still uses them internally. That is the library's own dependency, not this Step's code.

## Testing

- `go build ./...`, `go vet ./...` (only the pre-existing `t.Fatal`-from-goroutine warnings remain), and `go test ./...` all pass.
- `go mod tidy` / `go mod vendor` run; `go mod verify` is clean.


test/test.go
```diff
@@ -146,6 +145,9 @@ The Test Deploy directory has the following directory structure:
 			└── test-info.json
 */
 func ParseTestResults(testsRootDir string, useLegacyXCResultExtractionMethod bool, logger logV2.Logger) (results Results, err error) {
```
godrei: Can't we pass down the `pathModifier` and `pathChecker` created here: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/blob/65c6852f606312b0031ed08813f71aafa79465bd/main.go#L146
Similarly how to logger is passed down from the `main` func.

collect_files_to_deploy_integration_test.go
```diff
@@ -0,0 +1,105 @@
+// Integration tests that zip real files on disk: symlink and permission handling cannot be
```
ofalvai: Shouldn't this file be called `main_test.go`? I think the line between unit and integration tests are blurry, no need to mark this as an integration test (we are running this test as part of the regular tests).

collect_files_to_deploy_integration_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"archive/zip"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	loggerV2 "github.com/bitrise-io/go-utils/v2/log"
+	pathutil2 "github.com/bitrise-io/go-utils/v2/pathutil"
```
ofalvai: I don't see why these import aliases are needed, what was the conflict?

collect_files_to_deploy_integration_test.go
godrei: True, merged into main_test.go in https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/270/commits/cbf962b62f8af0f14b1a3b42a245e4a435fa9fec

collect_files_to_deploy_integration_test.go
godrei: Fixed the unnecessary import aliases across the codebase here: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/270/commits/cbf962b62f8af0f14b1a3b42a245e4a435fa9fec

ofalvai (approved):

PR #271, opened 2026-08-28
Migrate the log package to v2

godrei:
Stacked on top of #270. Review/merge that first.

## What

Thread a v2 `log.Logger` through the `uploaders` and `deployment` packages, replacing the v1 `go-utils/log` package-level functions with logger instances:

- **uploaders**: `createArtifact`, `UploadArtifact`, `finishArtifact` and `createSnapshot` now take a `logger` (the multipart helpers already did). Callers and tests updated.
- **deployment**: `Collector` and `ZipComparator` gain a `logger` field wired through their constructors (`NewCollector` / `NewZipComparator`); `main.go` and the deployment tests pass it in.
- **main.go**: drops the redundant v1 `SetEnableDebugLog` call (the v2 `EnableDebugLog` was already there) and switches the remaining `log.Printf` calls to the injected logger.

## Not in scope

The `androidartifact.Logger` adaptor in `uploaders/logger.go` still uses the v1 `log` package for its `RWarnf`-based APK/AAB parse warnings, so `go-utils` v1 stays in `go.mod` for now. That is migrated in the follow-up PR.

## Testing

`go build ./...`, `go vet ./...` (only the pre-existing `t.Fatal`-from-goroutine warnings remain) and `go test ./...` all pass.

### DIFF
--- a/uploaders/fileuploader.go
+++ b/uploaders/fileuploader.go
@@ -6,7 +6,7 @@ import (
 	"os"
 	"path/filepath"
 
-	logV2 "github.com/bitrise-io/go-utils/v2/log"
+	"github.com/bitrise-io/go-utils/v2/log"
 	"github.com/bitrise-steplib/steps-deploy-to-bitrise-io/deployment"
 )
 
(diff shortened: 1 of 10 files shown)


uploaders/fileuploader.go
```diff
@@ -6,7 +6,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/bitrise-io/go-utils/log"
+	logV2 "github.com/bitrise-io/go-utils/v2/log"
```
ofalvai: Is this really needed? What was the conflict?

godrei: Not needed, fixed here: https://github.com/bitrise-steplib/steps-deploy-to-bitrise-io/pull/271/commits/d7832a3e46cc51ab4439b64f230a7430443b5496

ofalvai (approved):

